PR #36, opened 2023-07-25
Fix broken link to Master-Detail Form example

llagerlof:


lovasoa:
Thank you, @llagerlof!

PR #71, opened 2023-08-23
Fix broken search in shell template

maaxleq:
The shell template was using a "target" attribute on the search <form> element, instead of an "action" attribute. The SQL script responsible for the search should be specified in an "action" attribute, otherwise the search does not work.


lovasoa:
Thanks for the fix!

PR #72, opened 2023-08-27
Fix ErrorWithStatus bubbling

edgrip:
Since 0.10.0 ErrorWithStatus show as html error rendering.
This leads to breaking Basic Auth mechanism for instance.
The fix considers every ErrorWithStatus must not show details as html.


lovasoa (approved):

lovasoa:
thanks @edgrip !

lovasoa:
I published your fix in 0.10.1 :+1:

PR #122, opened 2023-10-30
Updated Form component

Pieter3033:
Hi, @lovasoa 

**Changes:**
• Added 2 columns using bootstrap, to display tow input boxes next to each other.
• Adjusted the form button based on the new button component from SQLPage v0.14.
• Added a reset button, that essentially resets the form.

_Note_: The buttons are both top-level parameters.

**Problems:**
1. Struggling to adjust the width of the input boxes using bootstrap.

Form reference: 
image

SQLPage code:
image

### DIFF
--- a/sqlpage/templates/form.handlebars
+++ b/sqlpage/templates/form.handlebars
@@ -3,80 +3,106 @@
         {{#if title}}
             <h2 class="text-center mb-3">{{title}}</h2>
         {{/if}}
-        {{#each_row}}
-            {{#if (or (eq type "radio") (eq type "checkbox"))}}
-                <div class="form-selectgroup form-selectgroup-boxes d-flex flex-column my-1">
-                    <label class="form-selectgroup-item flex-fill">
-                        <input type="{{type}}" name="{{name}}" value="{{value}}" {{#if required}}required{{/if}} {{#if checked}}checked{{/if}} class="form-selectgroup-input">
-                        <div class="form-selectgroup-label d-flex align-items-center p-3">
-                            <div class="me-3">
-                                <span class="form-selectgroup-check"></span>
+        <div class="row">
+            {{#each_row}}
+            <div class="col-md-6 mb-2" >
+                {{#if (or (eq type "radio") (eq type "checkbox"))}}
+                    <div class="form-selectgroup form-selectgroup-boxes d-flex flex-column my-1">
+                        <label class="form-selectgroup-item flex-fill">
+                            <input type="{{type}}" name="{{name}}" value="{{value}}" {{#if required}}required{{/if}} {{#if checked}}checked{{/if}} class="form-selectgroup-input">
+                            <div class="form-selectgroup-label d-flex align-items-center p-3">
+                                <div class="me-3">
+                                    <span class="form-selectgroup-check"></span>
+                                </div>
+                                <div>
+                                    {{default label value}}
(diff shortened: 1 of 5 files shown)


sqlpage/templates/form.handlebars
```diff
@@ -3,80 +3,103 @@
...
-            {{#if (or (eq type "radio") (eq type "checkbox"))}}
-                <div class="form-selectgroup form-selectgroup-boxes d-flex flex-column my-1">
-                    <label class="form-selectgroup-item flex-fill">
-                        <input type="{{type}}" name="{{name}}" value="{{value}}" {{#if required}}required{{/if}} {{#if checked}}checked{{/if}} class="form-selectgroup-input">
-                        <div class="form-selectgroup-label d-flex align-items-center p-3">
-                            <div class="me-3">
-                                <span class="form-selectgroup-check"></span>
+        <div class="row">
+            {{#each_row}}
+            <div class="col-md-6 mb-2" >
```
lovasoa, line 8: I don't think we want to hardcode the `col-md-6` here. This would break existing forms. What I think we want is a new `columns` numeric row-level attribute that controls this CSS property.

sqlpage/templates/form.handlebars
```diff
@@ -3,80 +3,103 @@
...
+        {{/if}}
+        {{#if reset}}
+            <input class="btn 
+            {{#if reset_color}} btn-{{reset_color}} {{/if}}
+            {{#if reset_shape}} btn-{{reset_shape}} {{/if}}
+            {{#if reset_outline}} btn-outline-{{reset_outline}} {{/if}}
+            {{#if reset_size}} btn-{{reset_size}} {{/if}}"
+            type="reset" 
+            value="{{reset}}">
+        {{/if}}
```
lovasoa, line 103: Don't you think we should wrap that in some sort of container, maybe a flexbox ? Or just add a margin ? 
The way it is now, the two buttons feel too close to each other.

lovasoa (changes requested):
Wow, that looks great ! I highlighted a few changes that I think need to be made.

This also can't be merged without a corresponding update to the documentation: https://github.com/lovasoa/SQLpage/blob/main/examples/official-site/sqlpage/migrations/01_documentation.sql#L222-L328

You can make the changes in your branch, and they will be reflected in this PR automatically.

sqlpage/templates/form.handlebars
```diff
@@ -3,80 +3,103 @@
...
                         <small class="form-hint mt-0">{{description}}</small>
                     {{/if}}
                 </label>
             {{/if}}
-
+            </div>
         {{/each_row}}
-        <input class="btn btn-primary" type="submit" {{#if validate}}value="{{validate}}"{{/if}}>
+        </div>
+        {{#if validate}}
```
lovasoa, line 86: This would break all existing form that don't have a custom "validate" message.

lovasoa (approved):

lovasoa:
thanks @Pieter3033 ! your work will be included in the next version of sqlpage

Pieter3033:
Thanks @lovasoa, it feels I (as user) have more control over the form component now. 

I find it very interesting how you implemented the width of the columns with handlebars, I believe my problem was I never set the default of the width and it messed with the input-box.

I thank you though, because I am learning a-lot from your project. Keep it up man, and good luck.

PR #135, opened 2023-11-18
delegate statement preparation to sqlx

### LINKED ISSUE #100: Create and query in the same page
I was trying to create a table and query it from within the same SQL page.

Unfortunately, I was greeted with an error message:
`Failed to prepare SQL statement: 'SELECT * FROM zzmytable'`

Is there a workaround or an easy fix that I can apply?

The minimal working example triggering the above error message is as follows.

```sql
select 'shell' as component,
    'Create and query' as title,
    '/dbg-create-query.sql' as link;

select 'list' as component, 'Output' as title;

create table zzmytable as with t
       	(row)
	as (values 
	('A')
	) select * from t;
select * from zzmytable;
drop table zzmytable;
```

Initially, my aim was to create a temporary table and query that. However, not even permanent tables can be queried like that, as the above example shows. In the context of the psql client the above code is just fine. I would be grateful for any hints.

Thanks a lot!

lovasoa:
The logic of preparing statements and caching them for later reuse is now entirely delegated to the sql driver library (sqlx).
This simplifies the code and logic inside sqlpage itself.

 More importantly, statements are now prepared in a streaming fashion when a file is first loaded, instead of all at once, which allows referencing a temporary table created at the start of a file in a later statement in the same file.

 fixes #100


lovasoa:
@ipftalabua @nanawel : this pr should solve your respective problems, by allowing creating a temporary table and then referencing it from inside the same `.sql` file.

PR #162, opened 2023-12-29
Add tracking component

olivierauverlot:
Implementation of the tracking component (https://tabler.io/docs/components/tracking)

### DIFF
--- a/examples/official-site/sqlpage/migrations/28_tracking_component.sql
+++ b/examples/official-site/sqlpage/migrations/28_tracking_component.sql
@@ -48,6 +48,14 @@ VALUES (
     ),
     (
         'tracking',
+        'width',
+        'Width of the component, between 1 and 12.',
+        'NUMBER',
+        TRUE,
+        TRUE
+    ),
+    (
+        'tracking',
         'placement',
         'Position of the tooltip (e.g. top, bottom, right, left)',
         'TEXT',
@@ -56,15 +64,15 @@ VALUES (
     ),
     (
         'tracking',
-        'state',
-        'State of the tracked item (e.g. success, warning, danger)',
+        'color',
+        'Color of the tracked item (e.g. success, warning, danger)',
         'TEXT',
         FALSE,
         TRUE
     ),
     (
         'tracking',
-        'tooltip',
+        'title',
         'Description of the state.',
         'TEXT',
         FALSE,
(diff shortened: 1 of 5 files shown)


olivierauverlot:
Divider component added

lovasoa (changes requested):
Wow, these look great, thank you for the useful contribution !

Two small remarks:

 - To make it as easy as possible to switch between components and to memorize properties, we try to reuse property names as much as possible between components. In that spirit, I think the `state` property of the tracking component should be named `color` instead. One can add an example with `'danger' as color` to illustrate color name aliases. And maybe the tooltip should be named `title`, for consistency.
 - The spacing and sizing inside the tracking card seems a little off to me, when compared to tabler's example : 
   - image
   - image
   - do you think we could make it a little more compact, and maybe make the text stand out more ?
   - I think we should add a small vertical margin at the end of the component, as in other components. This looks more natural when used in combination with other components.

sqlpage/templates/tracking.handlebars
```diff
@@ -0,0 +1,20 @@
+<div class="card">
+  <div class="card-body">
+	 <div class="d-flex align-items-center">
+		<div class="subheader">{{title}}</div>
+	 </div>
+	<div class="d-flex align-items-baseline">
+		<div class="h1 mb-3 me-2">{{#if information}}{{information}}{{/if}}</div>
+	 </div>
+	 <div class="d-flex align-items-baseline">
+		<div class="mb-3 me-2">{{#if description}}{{description}}{{/if}}{{#if description_md}}{{{markdown description_md}}}{{/if}}</div>
```
lovasoa, line 10: You can use the [`remove-bottom-margin`](https://github.com/lovasoa/SQLpage/blob/main/sqlpage/sqlpage.css#L15) class here to prevent the double margin (from the div and from the `<p>`) when using markdown.

examples/official-site/sqlpage/migrations/28_tracking_component.sql
```diff
@@ -0,0 +1,94 @@
...
+    );
+-- Insert example(s) for the component
+INSERT INTO example(component, description, properties)
+VALUES      
+    (
+        'tracking', 
+        'An example of servers tracking component', 
+            JSON(
+                '[
+                {"component":"tracking","title":"Servers status","information":"60% are running","description_md":"Status of all **available servers** that actually running","placement":"top"},
```
lovasoa, line 81: I'm not sure I understand the example description. Maybe : 

```suggestion
                {"component":"tracking","title":"Servers status","information":"60% are running","description_md":"Status of all **currently running servers**","placement":"top"},
```

examples/official-site/sqlpage/migrations/28_tracking_component.sql
```diff
@@ -0,0 +1,94 @@
...
+        'TEXT',
+        FALSE,
+        FALSE
+    );
+-- Insert example(s) for the component
+INSERT INTO example(component, description, properties)
+VALUES      
+    (
+        'tracking', 
+        'An example of servers tracking component',
```
lovasoa, line 78: Maybe we can add a "simple" example with all optional fields omitted ?

olivierauverlot: done !

sqlpage/templates/tracking.handlebars
```diff
@@ -0,0 +1,25 @@
...
+	<div class="card-body">
+		<div class="d-flex align-items-center">
+			<div class="subheader">{{title}}</div>
+		</div>
+		<div class="d-flex align-items-baseline remove-bottom-margin">
+			<div class="h1 mb-3 me-2">{{#if information}}{{information}}{{/if}}</div>
+		</div>
+		<div class="d-flex align-items-baseline">
+			<div class="mb-3 me-2
+				{{#if description_md}} remove-bottom-margin">{{{markdown description_md}}}
```
lovasoa, line 11: I think you can simply apply  remove-bottom-margin regardless. If the div has no children (not in markdown mode), it will simply have no effect.

olivierauverlot: yes. I've simplified the template

lovasoa (approved):
Great, let's merge this ! This will be a nice christmas present for everyone :)

lovasoa:
Thanks again for your contribution, @olivierauverlot ! PRs adding useful components are always welcome !

lovasoa:
I'll try to make a new release soon that includes your work

PR #175, opened 2024-01-05
Support lazy-loaded content fragments in card component

lorefnon:
This PR enables the following: 

1. We can add an `?embed` query parameter in the url to skip the shell container
2. We can add a row level `embed=somepage.sql` parameter to lazy-load fragments of external content into cards
3. The js behavior is modified to support interactive components in lazy-loaded fragments

While 1 is useful in itself, for embedding externally generated html content within SQLPage rendered content, it is especially useful in conjunction with 2 because we can now render output of other pages including charts, maps etc. and embed them within cards.


lovasoa:
Wow, the results look great ! This is a useful feature, and seems easy enough to use. We will merge this. 
But I have two small concerns:
 - `embed` is a short keyword, there may be existing sqlpage applications out there using it as a URL parameter, which will make them render improperly. If we go with this design, then I would use something like `_sqlpage_embed`, and add somewhere to the documentation that `_sqlpage_*` url parameters are reserved.
 - manually implementing fragments implies quite some complexity to make everything work. Would it be a problem to simply use iframes and let the browser do the work ? Are there problems that I don't see ?

lorefnon:
Hi @lovasoa - thanks for taking a look. 

Wrt. first point I don't mind changing the param. 

Wrt. second - I don't believe iframes are an ideal solution. Because of the kind of strong sandboxing they offer, they are heavyweight and each frame creates its own dom tree, separate js context etc. 

The kind of use case I am currently prototyping has a few levels of drill down and for each level I'll have a grid of a dozen or so small charts in a page. So this means tabler css, apexcharts js etc. will be parsed and evaluated six times which is somewhat wasteful. 

Also any change in filters etc. in sqlpage currently results in a full page refresh, so these frames will be destroyed and created again and again for every click. 

I have worked with an old reporting solution in past that heavily used iframes and it was not a great experience for users. 

If you think that needing to manage fragments makes the js side of things more complex then one option is to use a library like [htmx](https://htmx.org) or [unpoly](https://unpoly.com) that makes swapping in dom fragments and interacting with them easier. 

But yeah, it could entirely be possible that I am stretching sqlpage a bit too far from its intended use case. Let me know if you think that this may cause issues for you wrt. long term maintenance overhead. Based on your response I'll reevaluate what works best for my dashboard.

lorefnon:
Also I don't mind adding support for frames in addition to inline html. Perhaps something like embed_mode="inline" vs embed_mode="iframe" - this could be useful for eg. if someone is embedded external content that does actually need sandboxing or style isolation.

lovasoa:
The argument of avoiding the overhead in case of a large number of frames is valid.
Indeed, having both options would be ideal, but this can be done in a second, separate, pull request.

I'm ready to merge this already if you just fix the CI error and rename the URL parameter.

lorefnon:
Ok, great. I have updated the query param, and the CI passes now.

examples/official-site/sqlpage/migrations/31_card_docs_update.sql
```diff
@@ -0,0 +1,33 @@
+DELETE FROM component WHERE name = 'card';
```
lovasoa, line 1: Ooopsie ! This removes all usage examples from the documentation.

I think you should re-add the examples from before here, and even add a new example illustrating the embed property.

If you need things to embed, you can put them in `examples/official-site/examples`

```sql

INSERT INTO example(component, description, properties) VALUES
    ('card', 'The most basic card', json('[{"component":"card"},{"description":"A"},{"description":"B"},{"description":"C"}]')),
    ('card', 'A card with a Markdown description',
            json('[{"component":"card", "columns": 2}, {"title":"A card with a Markdown description", "description_md": "This is a card with a **Markdown** description. \n\n'||
            'This is useful if you want to display a lot of text in the card, with many options for formatting, such as '||
            '\n - **bold**, \n - *italics*, \n - [links](index.sql), \n - etc."}]')),
    ('card', 'A beautiful card grid with bells and whistles, showing examples of SQLPage features.',
            json('[{"component":"card", "title":"Popular SQLPage features", "columns": 2},
            {"title": "Download as spreadsheet", "link": "?component=csv#component", "description": "Using the CSV component, you can download your data as a spreadsheet.", "icon":"file-plus", "color": "green", "footer_md": "SQLPage can both [read](?component=form#component) and [write](?component=csv#component) **CSV** files."},
            {"title": "Custom components", "link": "/custom_components.sql", "description": "If you know some HTML, you can create your own components for your application.", "icon":"code", "color": "orange", "footer_md": "You can look at the [source of the official components](https://github.com/lovasoa/SQLpage/tree/main/sqlpage/templates) for inspiration."}
    ]')),
    ('card', 'A gallery of images.',
        json('[
            {"component":"card", "title":"My favorite animals in pictures", "columns": 3},
            {"title": "Lynx", "description_md": "The **lynx** is a medium-sized **wild cat** native to Northern, Central and Eastern Europe to Central Asia and Siberia, the Tibetan Plateau and the Himalayas.", "top_image": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/d8/Lynx_lynx-4.JPG/640px-Lynx_lynx-4.JPG", "icon":"star" },
            {"title": "Squirrel", "description_md": "The **chipmunk** is a small, striped rodent of the family Sciuridae. Chipmunks are found in North America, with the exception of the Siberian chipmunk which is found primarily in Asia.", "top_image": "https://upload.wikimedia.org/wikipedia/commons/thumb/b/be/Tamias-rufus-001.jpg/640px-Tamias-rufus-001.jpg" },
            {"title": "Spider", "description_md": "The **jumping spider family** (_Salticidae_) contains more than 600 described genera and about *6000 described species*, making it the largest family of spiders with about 13% of all species.", "top_image": "https://upload.wikimedia.org/wikipedia/commons/thumb/a/ab/Jumping_spiders_%28Salticidae%29.jpg/640px-Jumping_spiders_%28Salticidae%29.jpg" }
        ]'));
```

lovasoa (changes requested):

examples/official-site/sqlpage/migrations/31_card_docs_update.sql
lovasoa: You can visualize the documentation locally just by running `cargo run` from `examples/official-site`

lorefnon: Ah sorry about that. Updated now.

lovasoa (approved):
Thanks again, this is a great addition to sqlpage

lovasoa:
I hadn't been careful, but it looks like this broke support for the `height` property of the chart component. All charts are now way too large.

lorefnon:
Yes,  sorry about that.  This fix is not ideal though.  I'll send a PR.

lovasoa:
thanks @lorefnon !

PR #178, opened 2024-01-07
New 'center' parameter to whether to center the tracking component.

olivierauverlot:


examples/official-site/sqlpage/migrations/32_tracking_update.sql
```diff
@@ -0,0 +1,17 @@
...
+    )
+VALUES (
+        'tracking',
+        'center',
+        'Whether to center the component.',
+        'BOOLEAN',
+        TRUE,
+        TRUE
+    );
+
```
lovasoa, line 17: Can we keep that in the existing tracking component documentation file ?

The database behind sql.ophir.dev is in-memory anyway and migrations are reapplied on every deploy.

lovasoa (changes requested):
I'd like to make the docs easier to edit by keeping them in one file per component. 

The change itself is good

examples/official-site/sqlpage/migrations/32_tracking_update.sql
olivierauverlot: ok, I will merge the two files :)

olivierauverlot:
The files are merged

lovasoa (approved):

lovasoa:
Thanks @olivierauverlot !

PR #179, opened 2024-01-07
New carousel component

olivierauverlot:


examples/official-site/sqlpage/migrations/33_carousel_component.sql
```diff
@@ -0,0 +1,150 @@
...
+        'name',
+        'Name of the carousel.',
+        'TEXT',
+        TRUE,
+        FALSE
+    ),
+    (
+        'carousel',
+        'title',
+        'Title of the carousel.',
```
lovasoa, line 28: The docs does not make it clear what the difference between name and title is.

examples/official-site/sqlpage/migrations/33_carousel_component.sql
```diff
@@ -0,0 +1,150 @@
...
+        'TEXT',
+        TRUE,
+        TRUE
+    ),
+    (
+        'carousel',
+        'indicators',
+        'Style of image indicators (e.g. square, dot).',
```
lovasoa, line 36: We should list possible values exhaustively here.

examples/official-site/sqlpage/migrations/33_carousel_component.sql
```diff
@@ -0,0 +1,150 @@
...
+        'carousel',
+        'title',
+        'Add caption to the slide.',
+        'TEXT',
+        FALSE,
+        TRUE
+    ),
+    (
+        'carousel',
+        'contents',
```
lovasoa, line 99: Can we call this description to be consistent with other components ?

sqlpage/templates/carousel.handlebars
```diff
@@ -0,0 +1,40 @@
...
+                {{#each_row}}
+                    <button type="button" data-bs-target="#{{../name}}" data-bs-slide-to="{{@row_index}}" {{#if (eq @row_index 0)}}class="active"{{/if}}></button>
+                    {{#delay}}
+                        {{flush_delayed}}
+                        <div class="carousel-item {{#if (eq @row_index 0)}}active{{/if}}">
+                            <img class="d-block w-100" alt="{{image}}" src="{{image}}" />
+                            {{#if title}}
+                            <div class="carousel-caption d-none d-md-block">
+                                <h5>{{title}}</h5>
+                                {{#if contents}}<p>{{contents}}</p>{{/if}}
```
lovasoa, line 19: Can we add the possibility to render markdown descriptions ?

Like other components, with `description` and `description_md`

lovasoa (changes requested):

examples/official-site/sqlpage/migrations/33_carousel_component.sql
olivierauverlot: I changed with 'An unique string to identify the caroussel component in the HTML page.' and 'A name to display at the top of the carousel.'

examples/official-site/sqlpage/migrations/33_carousel_component.sql
olivierauverlot: You're right. I think that "description" is more consistent.

olivierauverlot:
Requested changes done

examples/official-site/sqlpage/migrations/33_carousel_component.sql
```diff
@@ -0,0 +1,138 @@
...
+        top_level,
+        optional
+    )
+VALUES (
+        'carousel',
+        'name',
+        'An unique string to identify the caroussel component in the HTML page.',
+        'TEXT',
+        TRUE,
+        FALSE
```
lovasoa, line 23: We try to avoid having such requirements in sqlpage. We can do with it if there is no other choice, but it looks like this is an internal implementation technicality. Wouldn't it be better if the user didn't have to specify an id and make sure it is unique ? It is very easy to end up with duplicate identifiers and errors that are completely impossible to debug if you don't already understand html, javascript and the DOM.

Couldn't we auto-generate an id like `carousel_{{query_number}}` where query_number comes from https://github.com/lovasoa/SQLpage/blob/main/src/render.rs ?

lovasoa (changes requested):
Sorry for the double-review, but I didn't realize what `name` was.

examples/official-site/sqlpage/migrations/33_carousel_component.sql
olivierauverlot: Yes, it's possible to use an id like carousel_{{query_number}} but it seems to me not accessible in the template. If I understand the Rust code, the value of query_number is only initialized where an error is occured and only in development mode.

lovasoa: Yes render.rs needs to be updated to add query_number to the handlebars context.

lovasoa: You can look at how `row_index` is added to the context.
`query_number` needs to be passed all the way down to the `SplitTemplateRenderer`

lovasoa:
I reopened this is #214 and implemented the changes we talked about.

PR #182, opened 2024-01-10
Support id/class in built-in components

lorefnon:
Hello, currently some of the components support id/class like the form components. 

I'd like to propose that these be supported in all the built in components where it makes sense. 

I am primarily interested in id which is useful for anchor targeting (`<a href="#id">Some panel</a>`). It makes it possible to have links that scroll to a particular card or chart. For components that have a top level container, the id can be added to the container, and for those that don't have it can be added as a separate anchor.

class is nice to have for styling convenience. 

Hope this feature is acceptable. Happy to revise if you have any suggestions.


sqlpage/templates/card.handlebars
```diff
@@ -20,10 +21,10 @@
         row-cols-xl-5
     {{/if}}
     gx-2 gy-2
-    mt-1 mb-3">
+    mt-1 mb-3 {{class}}">
     {{#each_row}}
         <div class="col">
```
lovasoa, line 26: Wouldn't you like to allow giving individual cards an id ? This would allow linking to one card in particular.

sqlpage/templates/form.handlebars
```diff
@@ -1,8 +1,11 @@
 <form 
     {{#if id}}id="{{id}}"{{/if}}
-    class="my-3"
+    class="my-3 {{class}}"
     method="{{default method "post"}}"
-    {{#if action}}action="{{action}}"{{/if}}
+    {{#if action}}action="{{action}}"
+    {{else}}
+    {{#if id}}action="#{{id}}"{{/if}}
```
lovasoa, line 7: This behavior needs to be documented

sqlpage/templates/hero.handlebars
```diff
@@ -1,3 +1,4 @@
+{{#if id}}<a id="{{id}}" />{{/if}}
```
lovasoa, line 1: why not directly on the header element ?

lovasoa (changes requested):
Great, that sounds useful !

I'll merge, but all of the new attributes need to be documented first in https://github.com/lovasoa/SQLpage/blob/main/examples/official-site/sqlpage/migrations/01_documentation.sql (and other doc files in https://github.com/lovasoa/SQLpage/tree/main/examples/official-site/sqlpage/migrations )

lorefnon:
Hi @lovasoa - I have added the docs and updated as per above comments.

lovasoa (approved):
:confetti_ball:

PR #195, opened 2024-01-19
Prevent retention of cached object reference across async boundary

lorefnon:
[Context](https://github.com/lovasoa/SQLpage/issues/189#issuecomment-1900495003)

### DIFF
--- a/src/file_cache.rs
+++ b/src/file_cache.rs
@@ -4,6 +4,7 @@ use actix_web::http::StatusCode;
 use anyhow::Context;
 use async_trait::async_trait;
 use chrono::{DateTime, TimeZone, Utc};
+use dashmap::DashMap;
 use std::collections::HashMap;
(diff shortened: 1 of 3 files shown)


src/file_cache.rs
```diff
@@ -96,21 +96,25 @@ impl<T: AsyncFromStrWithState> FileCache<T> {
     }
 
     pub async fn get(&self, app_state: &AppState, path: &PathBuf) -> anyhow::Result<Arc<T>> {
-        log::trace!("Attempting to get from cache {:?}", path);
-        if let Some(cached) = self.cache.read().await.get(path) {
+        if let Some(cached) = self.cache.get(path) {
```
lovasoa, line 99: I think we just need to clone the result of `self.cache.get` before if..let here.

lovasoa (changes requested):

PR #216, opened 2024-02-04
Clarifies documentation site regarding the need to enclose http_header hyphenated names + typo fixes

matthewlarkin:
Hi @lovasoa, the documentation site needs to specify that hyphenated http_headers (Cache-Control, Content-Disposition, etc) _do require the use of enclosing quotes when used as aliases_.
```sql
-- custom htmx header
select 'http_header' as component,
    "/home.sql" as "HX-Redirect";
```
It was also somewhat unclear that we could declare custom headers; the examples only specified a handful of standard headers, which might imply that we're limited to those.

Also, found some related typos in the migration comments 🐛

### DIFF
--- a/examples/official-site/sqlpage/migrations/04_http_header.sql
+++ b/examples/official-site/sqlpage/migrations/04_http_header.sql
@@ -2,14 +2,20 @@
 INSERT INTO component (name, description, icon)
 VALUES (
         'http_header',
-        'An advanced component that can be used to create redirections, set a custom caching policy to your pages, or set any HTTP header.
+        'An advanced component that can be used to create redirections, set a custom caching
+            policy to your pages, and set any HTTP header, among other things.
         If you are a beginner, you probably don''t need this component.
-        When used, this component has to be the first component in the page, because once the page is sent to the browser, it is too late to change the headers.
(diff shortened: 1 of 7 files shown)


lovasoa:
The idiomatic syntax to include special characters in column names varies across databases. For instance, it will be `\`Cache-Control\`` in mysql and `[Cache-Control]` in Microsoft SQL Server.

lovasoa:
The correct place to include the fix to avoid generating invalid SQL in the documentation is `documentation.sql:

                                     CASE WHEN key LIKE '% %' THEN format('"%s"', replace(key, '"', '""')) ELSE key END as key,

We check whether the column name contains a space, we can test for hyphen too, and probably use [`format`](https://www.sqlite.org/printf.html#percentw) for escaping.

lovasoa:
I updated the documentation in https://github.com/lovasoa/SQLpage/commit/b44519a0b78e765d0bea3548f68b69a7d29ed6f8

Let me know if it can still be improved !

lovasoa:
https://sql.ophir.dev/documentation.sql?component=http_header#component

matthewlarkin:
Much clearer now! And neat that the official site data is just page caching and a bunch of migrations for an on-the-fly database 😎

PR #250, opened 2024-03-04
Form improvement : A unique identifier for the input fields

olivierauverlot:
A unique identifier for the input, which can then be used to select and manage the field with Javascript code. Usefull for advanced using as setting client side event listeners, interactive control of input field (disabled, visibility, read only, e.g.) and AJAX requests.


lovasoa (approved):
Very useful, thank you !

PR #251, opened 2024-03-04
Probe component

olivierauverlot:
A debug component that displays a value (constant, variable, parameter value, result of an SQL function, e.g.). Usefull to validate data flow or function behaviour. Each value (except NULL value) is surrounded by double quotes and followed by datatype (number, string, boolean, e.g.).


lovasoa:
Is there something that you don't like in the existing `debug` component?

olivierauverlot:
Yes, it may be necessary to modify the documentation for the debug component. The first sentence is too restrictive: "Display all the parameters passed to the component".

PR #252, opened 2024-03-04
New RSS component

olivierauverlot:
Produces a data flow in the RSS format. To use this component, you must first returning an HTTP header with the "application/rss+xml" content type (see http_header component). Next, you must use the shell-empty component to avoid that SQLPage generates HTML code. TIPS: If you don''t want change the page shell, you can also use the _sqlpage_embed URL parameter.


lovasoa:
I made a few changes, let me know what you think.

We still need to provide a RFC-822 date formater to the templates to make this work.

olivierauverlot:
Yes, perfect. It's ok for me

PR #255, opened 2024-03-06
A low level component to create a dedicated DOM area

olivierauverlot:
A low level component to create a dedicated DOM area to dynamically add, modify or delete elements. To this, you can use Vanilla JavaScript or using dedicated framework as JQuery. The container component is built from a DIV element. An identifier is mandatory to access the container content with the method getElementById() for example. This component is useful for inserting the result of an AJAX request in the page.

### DIFF
--- a/examples/official-site/sqlpage/migrations/38_container.sql
+++ b/examples/official-site/sqlpage/migrations/38_container.sql
@@ -1,0 +1,41 @@
+-- Documentation for the container component
+INSERT INTO component (name, description, icon, introduced_in_version) VALUES (
+    'container',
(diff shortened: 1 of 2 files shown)


lovasoa (changes requested):
Hey ! This is nice, but I don't think this belongs to the standard library of sqlpage... Maybe we should write an example in the examples directory that illustrates how to use a custom component and interact with it in javascript ?

olivierauverlot:
Are you sure? It's a current usage to implement this type of component in a web application. Adding this component avoid to reinvent the wheel for each project. At a semantic and structural levels, it's a good thing to clearly identify a component dedicated to this task. I think that it's better to propose a prepared solution rather than forcing the user to tinker with the text component.

PR #257, opened 2024-03-06
The image component embeds an image into the page.

olivierauverlot:


olivierauverlot:
Ok, I will try to do that

PR #265, opened 2024-03-14
Added a small draft sidebar template

srcarvalho12:
How to use:
SELECT
    'shell' As component,
    'Clientes' As title,
    TRUE As sidebar,
    
I had problems with "scrollbar-color", as it has a predefined style and I would like to use transparent, for example:
scrollbar-color: rgba(var(--tblr-scrollbar-color, var(--tblr-body-color-rgb)), .16) transparent;


lovasoa:
Great! Do you want to try implementing the sidebar + navbar combo ?

srcarvalho12:
I will improve in the next few days

PR #298, opened 2024-04-30
Adds `body_class` for form component

matthewlarkin:
Hey @lovasoa, what do you think of this tiny form component enhancement?

Allows adding classes to the fieldset element that contains the body of the form. Useful for things like inner padding, border adjustments, etc.

Would love to see something like this in the official component, and I can update the relevant documentation if this looks good to you!

### DIFF
--- a/sqlpage/templates/form.handlebars
+++ b/sqlpage/templates/form.handlebars
@@ -7,7 +7,7 @@
     {{#if id}}action="#{{id}}"{{/if}}
     {{/if}}
 >
-    <fieldset class="form-fieldset">
+    <fieldset class="form-fieldset {{body_class}}">
         {{#if title}}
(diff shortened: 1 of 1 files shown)


lovasoa:
Hi! 

That would be okay with me, but I'm not completely sure I understand the why. 

Is there something preventing the usage of the existing class attribute, and then styling it with

```css
.my_class fieldset {
  /*...*/
}
```

?

matthewlarkin:
Ah yes @lovasoa, thank you for helping me not to overthink it! I will remember to just use CSS next time 🤦‍♂️ 😄

Regardless, `body_class` would help the folks who just want to write SQL statements, if you think it's still worth it to add it.

Thanks as always!

PR #321, opened 2024-05-15
Add favicon and manifest properties to shell.handlebars

### LINKED ISSUE #317: Support custom favicon location and manifest.json
I like to provide a PWA for mobile devices.

Currently, no manifest.json is served at all (requires a meta link in HTML header as far as I know).

Similarly, the favicon is served, but the location is not configurable (relies on browser looking for domain.com/favicon.ico).

I'm not able to provide a PWA without the meta link in the shell component.

I guess I can work around with a custom shell component, but for now I forego the PWA.

The favicon I am also not serving, as it affects my entire domain root favicon.

**Describe the solution you'd like**

We should:
* serve manifest.json from root or site-prefix if exists (default meta link) - this could default to off
* serve favicon from root or site-prefix (add meta link)
* allow user to override defaults (set custom location for manifest.json and favicon.ico)

**Describe alternatives you've considered**

I tried routing the favicon via my reverse proxy, but it affects the whole domain. I need the meta link to tell the browser where to look other than the domain root.

These commits show what I would like (the <link> tag in the HTML <head>):

* https://github.com/lovasoa/SQLpage/pull/311/commits/e09e1dbb5a084854fab1c430522b138b5f6d870e
* https://github.com/lovasoa/SQLpage/pull/311/commits/97d9a3c250f125d829d82141766c88b0d66c260f

But we need to implement it through the shell component.

djyotta:
Resolves #317


djyotta:
@lovasoa Is the component documentation in a different codebase? I couldn't find it...

lovasoa:
The documentation is here: https://github.com/lovasoa/SQLpage/blob/b2f646aac0924c066fc0e942d6536e460d64c9bb/examples/official-site/sqlpage/migrations/01_documentation.sql#L761

You can edit the existing migration, since the official website's database is a volatile in-memory sqlite.

djyotta:
Cool! I spun up the documentation - not sure how I missed that. I should have grep'd for it.

Docs look ok to me. I tried to be restrict verbosity to be similar to the neighbouring cards.

examples/official-site/sqlpage/migrations/01_documentation.sql
```diff
@@ -739,6 +739,10 @@ and `shell.json` would be placed at the website''s root and contain the followin
 INSERT INTO component(name, icon, description) VALUES
     ('shell', 'layout-navbar', 'Personalize the "shell" surrounding your page contents. Used to set properties for the entire page.');
 
+INSERT INTO parameter(component, name, description_md, type, top_level, optional) SELECT 'shell', * FROM (VALUES
+    ('favicon', 'The URL of the icon the web browser should display in bookmarks and tabs. This property is particularly useful if multiple sites are hosted on the same domain with different [``site_prefix``](https://github.com/lovasoa/SQLpage/blob/main/configuration.md#configuring-sqlpage).', 'URL', TRUE, TRUE),
```
lovasoa, line 743: I think we render the descriptions here as text, not markdown. But we have markdown and syntax highlighting in the examples below, where we can be a little bit more verbose

lovasoa (changes requested):

lovasoa (approved):

PR #323, opened 2024-05-18
Support for listening on UNIX sockets

vlasky:
This specifies a path to a UNIX socket file to listen on instead of the TCP port.

If specified, SQLPage will accept HTTP connections only on this socket and not on any TCP port. This option is mutually exclusive with the `listen_on` and `port` options.

Useful when running SQLpage behind a proxy server like Nginx, as the overhead for communication using Unix Domain Sockets is less than when using the TCP stack.


lovasoa:
Great! Could you please also update our [nginx.sql](https://sql.ophir.dev/your-first-sql-website/nginx.sql) which explains how to setup SQLPage behind nginx ?

vlasky:
There are some other enhancements I want to make to this code so I will do that too.

src/webserver/http.rs
```diff
@@ -600,3 +606,25 @@ fn bind_error(e: std::io::Error, listen_on: std::net::SocketAddr) -> anyhow::Err
     };
     anyhow::anyhow!(e).context(ctx)
 }
+
+fn bind_uds_error(e: std::io::Error, unix_socket: &PathBuf) -> anyhow::Error {
+    let ctx = match e.kind() {
+        std::io::ErrorKind::AddrInUse => format!(
```
lovasoa, line 612: Can this error be returned when binding to a socket?

src/webserver/http.rs
```diff
@@ -600,3 +606,25 @@ fn bind_error(e: std::io::Error, listen_on: std::net::SocketAddr) -> anyhow::Err
...
+            "Another program is already using the UNIX socket {:?}. \
+            You can either stop that program or change the socket path in the configuration file.",
+            unix_socket,
+        ),
+        std::io::ErrorKind::PermissionDenied => format!(
+            "You do not have permission to bind to the UNIX socket {:?}. \
+            You can change the socket path in the configuration file or check the permissions.",
+            unix_socket,
+        ),
+        std::io::ErrorKind::AddrNotAvailable => format!(
```
lovasoa, line 622: This can't be returned when binding to a socket either, can it?

src/webserver/http.rs
lovasoa: Could you test the various error cases locally to check what ioerrors are actually returned?

lovasoa:
It's merged, thank you @vlasky !

PR #327, opened 2024-05-22
Adds SQL Migrations Introductory Article

matthewlarkin:
Hi @lovasoa,

Here's a SQL migrations introductory article for SQLPage! Let me know if you think it could use any refining. I image you know of some good places in the documentation to link to this from once it's ready to publish.

Thanks for the invite to contribute,

Matthew

### DIFF
--- a/examples/official-site/migrations.sql
+++ b/examples/official-site/migrations.sql
@@ -1,0 +1,153 @@
+select 'http_header' as component,
+	'public, max-age=300, stale-while-revalidate=3600, stale-if-error=86400' as "Cache-Control";
+
+select 'dynamic' as component, properties FROM example WHERE component = 'shell' LIMIT 1;
+
+-- Article by Matthew Larkin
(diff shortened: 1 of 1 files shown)


lovasoa:
Great, thank you very much @matthewlarkin ! That's better than anything I could have written. I'll add the links.

examples/official-site/migrations.sql
```diff
@@ -0,0 +1,153 @@
...
+	first_name not null,
+	last_name not null,
+	middle_name,
+	email not null unique,
+	phone,
+	rewards_level integer check(rewards_level between 1 and 20),
+	password_hash not null
+);
+
+insert into users_temp select * from users;
```
lovasoa, line 114: Maybe it would be a better practice to indicate the column names here?

matthewlarkin: I agree, just updated 👍

PR #348, opened 2024-05-29
Update button.handlebars adding after-icon

amrutadotorg:
{{#if after_icon}}: This new block checks if the after_icon parameter is provided. If it is, it renders the icon after the text. In this way, the template will support both an icon before and after the text, depending on which parameters (icon or after_icon) are provided.

### DIFF
--- a/sqlpage/templates/button.handlebars
+++ b/sqlpage/templates/button.handlebars
@@ -17,6 +17,9 @@
             <span class="me-1">{{~icon_img icon~}}</span>
         {{~/if~}}
         {{~title~}}
+        {{~#if after_icon~}}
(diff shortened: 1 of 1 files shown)


lovasoa:
Great, thanks for the pr ! Can you also, in the same pull request, add the new item to the documentation: https://github.com/lovasoa/SQLpage/blob/main/examples/official-site/sqlpage/migrations/18_button.sql ?

Don't close this PR, just amend it.

lovasoa:
@amrutadotorg : don't worry, I did it myself in https://github.com/lovasoa/SQLpage/commit/e5a964928e390c2eb4ffa1e45ae8b1e574e5b645 . Thanks for your contribution !

PR #349, opened 2024-05-29
Example of light/dark toggle switch

lyderic:
This is a very simple way to toggle between light and dark theme, easy to implement on any SQLpage site


lovasoa (approved):
Thanks for the useful contribution, @lyderic !

lovasoa:
I added a reference to your example in the main project's README to let people discover it more easily.

PR #355, opened 2024-05-30
Update card.handlebars. Add the style parameter to the iframe in embed

amrutadotorg:
This modification ensures that the style attribute is added to the iframe and div elements if the style parameter is provided. If style is not provided, the attribute is omitted.

### DIFF
--- a/sqlpage/templates/card.handlebars
+++ b/sqlpage/templates/card.handlebars
@@ -48,6 +48,7 @@
                             {{~#if height}} height="{{height}}"{{/if~}}
                             {{~#if allow}} allow="{{allow}}"{{/if~}}
                             {{~#if sandbox}} sandbox="{{sandbox}}"{{/if~}}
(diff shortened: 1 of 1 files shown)


sqlpage/templates/card.handlebars
```diff
@@ -66,7 +67,6 @@
                             <a href="{{footer_link}}">{{footer}}</a>
                     {{else}}
                             {{footer}}
-                    {{/if}}
```
lovasoa, line 69: Looks like you unintentionally removed this

lovasoa (changes requested):
Thanks! Can you also update the documentation here: https://github.com/lovasoa/SQLpage/blob/main/examples/official-site/sqlpage/migrations/31_card_docs_update.sql#L25 ?

amrutadotorg:
Done here #365

PR #357, opened 2024-05-30
 faster run_sql with shared database connection

lovasoa:
see https://github.com/lovasoa/SQLpage/issues/338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
  - new `tooltip` property in the button component.
  - New `search_value` property in the shell component.
  - Fixed a display issue in the hero component when the button text is long and the viewport is narrow.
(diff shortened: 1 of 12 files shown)


lovasoa:
@vlasky @guspower : this ended up being a larger change than expected... Would any of you be available to test this on your applications before I release the change for everyone ? I would like to make sure the change does not have any unexpected consequences.

guspower:
@lovasoa yes I can take it for a spin tomorrow and let you know.

guspower:
FYI I cloned `better_run_sql_management` yesterday and ran the release build against a reasonable size app - no issues functionally. I was going to cook up a performance test to compare against the previous release but I see you've merged in so I'll hold off for now.

lovasoa:
Great to hear it works. And I'd actually would be very interested in the numbers you get on your application in a performance test.

guspower:
Me too! First I am going to set up a functional check using `spider` and then figure out how to run a perf test (which is always a pain :) ) I doubt I'll get one running this week due to workload, but when I do I will share.

PR #359, opened 2024-05-31
the search query parameter is used as the value

amrutadotorg:
Search Input Field ID:

Added id="search-input" to the search input field for easier identification by the script.

JavaScript Code:

Function getQueryParams: This function parses the query string from the URL and returns an object containing key-value pairs of the parameters.
DOMContentLoaded Event Listener: This ensures that the script runs only after the entire page (DOM) has been loaded.

Setting the Input Value: If the search parameter is present in the URL, the script sets the value of the search input field to the value of this parameter.

### DIFF
--- a/sqlpage/templates/shell.handlebars
+++ b/sqlpage/templates/shell.handlebars
@@ -10,13 +10,13 @@
         <link rel="manifest" href="{{manifest}}">
     {{/if}}
     <link rel="stylesheet" href="{{static_path 'sqlpage.css'}}">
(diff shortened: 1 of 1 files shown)


lovasoa:
Hello! I'm not sure about serving and executing this script all the time, for everyone. Ideally, the server would serve the input element with a value attribute preset to whatever we want. The easiest is to set value="{{search_value}}" and let the programmer select `$search as search_value` themselves.

lovasoa:
Closing in favor of #362

PR #370, opened 2024-06-02
add uploaded_file_name function

### LINKED ISSUE #333: Preserve file name on file upload
**_What are you building with SQLPage ?_**

I have a suite of apps I use to help me transfer data between devices.
* It's not for sharing secrets, it's for convenience only. Not recommended to paste and private info... 

ie, I can paste something from the clip board on my PC to the web form, and then scan the QR code on my phone.
OR, on a device without a camera, the codes are 5 base 32 digits (yes, I'm not worried about collisions) so it's easy to remember and type.

Here is the tools:
* https://shandan.one/clip - paste text to share
* https://shandan.one/goto - tiny URL
* https://shandan.one/upload - upload file to share

I formerly used python + bottle and rolled my own UI. This is fine.

I persisted clip/link/file to disk by naming it by it's base32 hash. Lots of small files. This is fine.

On finding out about SQLPage, I'm converting these so my website has a similar look and feel. Also want to get rid of the horrific python code I wrote.

It also make sense to store this stuff in db so it's easier to cleanup old data and apply size limits (had a bash script to do it, but ... yeah not comfortable with it).

**_What is your problem ? A description of the problem, not the solution you are proposing._**

The drawback with the SQLPage version is that I can not preserve the file name on upload.

My old UI, I was preserving the file name supplied by the browser:
```
-----------------------------4053620313313727090337127195
Content-Disposition: form-data; name="action"

Upload
-----------------------------4053620313313727090337127195
Content-Disposition: form-data; name="content"; filename="match.sh"
Content-Type: application/x-shellscript

#!/bin/bash
grep -ni 'import .*gnu' "$@"

-----------------------------4053620313313727090337127195--
```
The `Content-Disposition` header includes the filename (at least on Brave and Firefox).

_**What are you currently doing ? Since your solution is not implemented in SQLPage currently, what are you doing instead ?**_

There doesn't seem to be a way to get the filename out of the form data on SQLPage.

For now, I'm using the base32 code as the file name - at least this is not random and will be consistent on each download and also relates back to the code used to share the file.

_**Describe the solution you'd like**_

I'd like to preserve the file name on upload, so that when users download the file, the don't get a random or meaningless name.

A SQLPage function like [uploaded_file_mime_type](https://sql.ophir.dev/functions.sql?function=uploaded_file_mime_type#function) would be good. Maybe call it `uploaded_file_name`. Can return NULL if user-agent doesn't provide it.

djyotta:
Closes #333 

Tested with
```sql
SET fname = sqlpage.uploaded_file_name('f');

SELECT 'debug' AS component;
SELECT $fname AS fname;

SELECT 'form' AS component, 'post' AS method;
SELECT 'file' AS type, 'f' AS name, '' AS label;
```

Documentation tbd

### DIFF
--- a/examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
+++ b/examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
@@ -137,69 +137,16 @@ insert into uploaded_files (fname, content, uploaded) values (
 
 > *Note*: Data URLs are larger than the original file, so it is not recommended to use them for large files.
 
-There is not currently support in SQLPage to download files, but a web server with access to the database
-can provide the download with the original file name.
+### Downloading the uploaded files
 
-Here is an example of a simple web server using `bottle` (`python`):
-```python
-from time import strftime, gmtime
-from sqlite3 import connect
-from datetime import datetime, timezone
-from bottle import route, abort, HTTPResponse
-from itertools import chain
-from base64 import b64decode
-
-def parse_data_uri(content):
-  _, *media, data = chain.from_iterable(map(
-      lambda x: x.split(",", 1), content.split(":", 1)
-  ))
-  media = media and media[0]
-  mimetype, *params, encoding = media.split(";")
-  if "=" in encoding:
-    params.append(encoding)
-    encoding = None
-  
-  return {
-    "mimetype": mimetype,
-    "params": dict(map(lambda x: x.split("="), params)),
-    "encoding": encoding,
-    "data": data,
-  }
-
-@route(f"/upload/<fname>", method="GET")
-def get_upload(fname):
-  con = connect("sqlpage.db")
-  content, uploaded = (None, None)
-  try:
-    content, uploaded = con.cursor().execute(f"""
-SELECT content, uploaded FROM uploaded_files WHERE fname = ''{fname}'' LIMIT 1;
-""").fetchall()[0]
-  finally:
-    con.close()
-
-  uploaded = datetime.strptime(uploaded, "%Y-%m-%d %H:%M:%S").replace(tzinfo=timezone.utc).timestamp()
-  data = parse_data_uri(content)
-  mimetype = data["mimetype"]
-  charset = data["params"].get("charset", None)
-
-  if data["encoding"] == "base64":
-    abort(500, f"""unsupported encoding: {data["encoding"]}""")
-
-  content = b64decode(data["data"] + "==")
-
-  if mimetype:
-    if mimetype[:5] == "text/" and charset and "charset" not in mimetype:
-        mimetype += "; charset=%s" % charset
-
-  headers = {
-    "Content-Length": len(content),
-    "Content-Disposition": f"attachment; filename=\"{fname}\"",
-    "Content-Encoding": "application/octet-stream",
-    "Content-Type": mimetype,
-    "Last-Modified": strftime("%a, %d %b %Y %H:%M:%S GMT", gmtime(uploaded)),
-  }
-  return HTTPResponse(content, **headers)
+The file can be downloaded by clicking a link like this:
+```sql
+select 'text' as component;
+select '<a download="'||fname||'" href="'||content||'">Click to download '||fname||'</a>' from uploaded_files
+where fname = $file_name as html;
 ```
+
+> *Note*: because the file is ecoded as a data uri, the file is transferred to the client whether or not the link is clicked
 
 ### Large files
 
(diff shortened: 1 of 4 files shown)


lovasoa (changes requested):
Great, thank you @djyotta !

Can you document the new function in https://github.com/lovasoa/SQLpage/tree/main/examples/official-site/sqlpage/migrations and add a test for it like https://github.com/lovasoa/SQLpage/blob/main/tests/index.rs#L263-L285 ?

djyotta:
Test and documentation added

examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
```diff
@@ -105,6 +105,108 @@ select ''redirect'' as component,
...
+  CURRENT_TIMESTAMP
+);
+```
+
+> *Note*: Data URLs are larger than the original file, so it is not recommended to use them for large files.
+
+There is not currently support in SQLPage to download files, but a web server with access to the database
+can provide the download with the original file name.
+
+Here is an example of a simple web server using `bottle` (`python`):
```
lovasoa, line 143: Why?

examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
```diff
@@ -105,6 +105,108 @@ select ''redirect'' as component,
...
+insert into uploaded_files (fname, content, uploaded) values (
+  sqlpage.uploaded_file_name(''myfile''),
+  sqlpage.read_file_as_data_url(sqlpage.uploaded_file_path(''myfile'')),
+  CURRENT_TIMESTAMP
+);
+```
+
+> *Note*: Data URLs are larger than the original file, so it is not recommended to use them for large files.
+
+There is not currently support in SQLPage to download files, but a web server with access to the database
```
lovasoa, line 140: Why do you say this? One can easily make SQLPage download the file for the user directly, with a button that contains the data url

lovasoa (changes requested):

lovasoa:
I don't think we can expect the SQLPage documentation readers to be familiar with python

examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
djyotta: I'll test and document this approach instead.

examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
djyotta: Because I didn't think of the data uri as an option for download. :)

djyotta:
> I don't think we can expect the SQLPage documentation readers to be familiar with python

I've modified the docs to demonstrate download via data uri instead of the python example.

lovasoa (changes requested):

examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
```diff
@@ -105,6 +105,57 @@ select ''redirect'' as component,
...
+### Downloading the uploaded files
+
+The file can be downloaded by clicking a link like this:
+```sql
+select 'text' as component, (
+  select '<a download="'||fname||'" href="'||content||'">Click to download '||fname||'</a>' from uploaded_files
+  where fname = $file_name
+  limit 1
+) as html;
+```
```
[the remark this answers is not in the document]
djyotta: Unfortunately, that doesn't download the file. I tested with an image, and the browser just displays it. Also there is no way for the browser to know the file name. Apparently, there has been some support for including the file name as a media parameter in the data uri, but seems the "download" attribute is the most supported way.

I guess we need some handlebars to make a download component...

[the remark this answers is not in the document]
lovasoa: You are right, if the file can be displayed in the browser directly, it will (and can be saved to disk from there). 

We should add a new `download` row-level attribute to the button component. But we can do that in a separate pull request. 

For the moment, can you apply the change I suggest above to the example? You can change it to add the download attribute in the next pr.

examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
```diff
@@ -105,6 +105,54 @@ select ''redirect'' as component,
...
+);
+```
+
+> *Note*: Data URLs are larger than the original file, so it is not recommended to use them for large files.
+
+### Downloading the uploaded files
+
+The file can be downloaded by clicking a link like this:
+```sql
+select 'button' as component;
```
lovasoa, line 144:
```suggestion
select ''button'' as component;
```

lovasoa (commented):
sorry, my mistake

examples/official-site/sqlpage/migrations/23_uploaded_file_functions.sql
djyotta: Ah I forgot to test the docs...

lovasoa (approved):

lovasoa:
Merged! Thank you again, @djyotta !

PR #393, opened 2024-06-11
adding image_og property to the shell component 

amrutadotorg:


examples/official-site/sqlpage/migrations/01_documentation.sql
```diff
@@ -764,6 +764,7 @@ INSERT INTO parameter(component, name, description, type, top_level, optional) S
     ('javascript', 'The URL of a Javascript file to load and execute on the page.', 'URL', TRUE, TRUE),
     ('rss', 'The URL of an RSS feed to display in the top navigation bar. You can use the rss component to generate the field.', 'URL', TRUE, TRUE),
     ('image', 'The URL of an image to display next to the page title.', 'URL', TRUE, TRUE),
+    ('image_og', 'The URL of an image to display in the Open Graph metadata.', 'URL', TRUE, TRUE),
```
lovasoa, line 767: Less technical users will probably not know what "Open Graph metadata" means. We could explain that it's the preview image that will appear if the page is shared on social media.

sqlpage/templates/shell.handlebars
```diff
@@ -43,6 +43,10 @@
         <link rel="alternate" type="application/rss+xml" title="{{title}}" href="{{rss}}">
     {{/if}}
     <meta name="generator" content="SQLPage"/>
+
+    {{#if image_og}}
```
lovasoa, line 47: Can we rename it to something like `social_image` ? Something that will be easier to understand by someone who hasn't heard of "open graph"

examples/official-site/sqlpage/migrations/01_documentation.sql
lovasoa: We should still mention open graph, but add the explanation

lovasoa (changes requested):

lovasoa:
Thanks for the pr ! Can you make the changes above ?

amrutadotorg (commented):
changed, I hope it is ok now

lovasoa (approved):

lovasoa:
thanks !

PR #405, opened 2024-06-14
adds semantic classes, ids, and tags

matthewlarkin:
Adds semantic classes, ids, and tag names

Should give us more semantic CSS styling control. Example, `body` element theming, [à la Shoelace](https://shoelace.style/getting-started/themes):
```sql
select 'shell' as component,
    'sl-theme-dark' as body_class,
    [...]
```

- - -

⚠️ The new `<footer>` and `<header>` tags may break styling that relied on css selections such as `body > div:last-of-type {...}` prior to semantic tagging.

### DIFF
--- a/sqlpage/templates/shell.handlebars
+++ b/sqlpage/templates/shell.handlebars
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{language}}" style="font-size: {{default font_size 18}}px" {{#if html_class}}class="{{html_class}}"{{/if}}>
+<html lang="{{language}}" style="font-size: {{default font_size 18}}px" {{#if class}}class="{{class}}"{{/if}}>
 <head>
     <meta charset="utf-8"/>
     <title>{{default title "SQLPage"}}</title>
--- a/sqlpage/templates/shell.handlebars
+++ b/sqlpage/templates/shell.handlebars
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{language}}" style="font-size: {{default font_size 18}}px">
+<html lang="{{language}}" style="font-size: {{default font_size 18}}px" {{#if html_class}}class="{{html_class}}"{{/if}}>
 <head>
     <meta charset="utf-8"/>
     <title>{{default title "SQLPage"}}</title>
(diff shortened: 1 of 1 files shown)


sqlpage/templates/shell.handlebars
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{language}}" style="font-size: {{default font_size 18}}px">
+<html lang="{{language}}" style="font-size: {{default font_size 18}}px" {{#if html_class}}class="{{html_class}}"{{/if}}>
```
lovasoa, line 2: We call it simply `class` everywhere else. Can you rename it ?

sqlpage/templates/shell.handlebars
```diff
@@ -49,76 +49,79 @@
...
+					<button class="navbar-toggler" type="button" data-bs-toggle="collapse"
+						data-bs-target="#navbar-menu" aria-controls="navbar-menu" aria-expanded="false"
+						aria-label="Toggle navigation">
+						<span class="navbar-toggler-icon"></span>
+					</button>
+					<div class="collapse navbar-collapse" id="navbar-menu">
+						<ul class="navbar-nav ms-auto">
+							{{#each (to_array menu_item)}}
+								{{#if (eq (typeof this) 'object')}}
+									<li class="nav-item{{#if this.submenu}} dropdown{{/if}}">
```
lovasoa, line 77: The identation change makes it hard to see the diff here

sqlpage/templates/shell.handlebars
```diff
@@ -49,76 +49,79 @@
...
-                        <form class="d-flex" role="search" action="{{search_target}}">
-                            <input class="form-control me-2" type="search" placeholder="Search" aria-label="Search"
-                                   name="search" value="{{search_value}}">
-                            <button class="btn btn-outline-success" type="submit">Search</button>
-                        </form>
-                    {{/if}}
-                </div>
-            </div>
-        </nav>
+		<header id="sqlpage_header">
```
lovasoa, line 55: What's the objective of this header element if it just contains a nav ?

sqlpage/templates/shell.handlebars
```diff
@@ -49,76 +49,79 @@
...
-<div class="w-100 text-center fs-6 my-2 text-secondary">
-    {{#if footer}}
-        {{{markdown footer}}}
-    {{else}}
-        <!-- You can change this footer using the 'footer' parameter of the 'shell' component -->
-        Built with <a class="text-reset" href="https://sql.ophir.dev" title="SQLPage v{{buildinfo 'CARGO_PKG_VERSION'}}">SQLPage</a>
-    {{/if}}
-</div>
+<footer class="w-100 text-center fs-6 my-2 text-secondary" id="sqlpage_footer">
+	<div class="sqlpage_footer_content">
```
lovasoa, line 117: Same as before, I'm not a huge fan of adding nesting if there isn't a good reason. Do we need footer + div, or can we go with just a footer ?

matthewlarkin: Regarding these two 👆: mostly to include semantics (`header` and `footer`) but also to provide _some_ structure for those who want to shape the layout a bit with CSS, if desired.

I'm not a fan of meaningless divs either, but I do like _some_ internal structure (as of now there is none). Having at least one internal div gives us a lot more flexibility for CSS layout, backgrounds, margin/padding, etc, without making a noticeable different if we don't use CSS (since the blocks collapse if not styled).

What do you think?

sqlpage/templates/shell.handlebars
matthewlarkin: Oh shoot, I'll fix that 😄 My editor is set to huge indents, and that must have slipped in here.

sqlpage/templates/shell.handlebars
matthewlarkin: Yes! I was considering a `body_class` too (which led to the distinction), but I later took that away. Probably not needed.

The point of having a class on the `<html>` at all I think is to assist with logical class names (such as the Shoelace custom theming I pointed to earlier) that CSS can latch on to.

Just `class` should be fine; `<html>` is the ultimate shell 👌

sqlpage/templates/shell.handlebars
lovasoa: Do you have an example of something you wanted to do that is harder without the internal div ?

lovasoa:
Thanks !

PR #417, opened 2024-06-17
Code mistake in 25_read_file_as_text.sql

pchemguy:
"text" property is specified instead of "contents_md"


lovasoa (approved):
Thanks for fixing this !

PR #422, opened 2024-06-18
Added "overflow" attribute to the "table" component

pchemguy:
The "table" component template will now set the width property of the table component div to "fit-content" if the table defines a boolean "overflow" attribute set to TRUE.

This feature is useful for "wide" tables, as it prevents columns clipping on the right, enabling horizontal scrolling via the browser scrollbar.


lovasoa (changes requested):
Thanks ! This also needs to be documented in https://github.com/lovasoa/SQLpage/blob/f49addfdaaf8ddb5219a1e0681d2858fb10313f3/examples/official-site/sqlpage/migrations/01_documentation.sql#L595

lovasoa (approved):

PR #423, opened 2024-06-18
Added "narrow" attribute to the "button" component

pchemguy:
Sets .btn-icon class, which trims horizontal padding.
See https://tabler.io/docs/components/buttons#icon-buttons


lovasoa (approved):
great !

PR #427, opened 2024-06-19
Added "parameter_type" table and associated FK

pchemguy:
The "parameter"("type") field contained two variants of BOOLEAN name (the other one being 'BOOL') due to lack of natural "enum" constraint. Added parameter_type table and associated foreign key to ensure consistency. Also replaced all "BOOL" instances with "BOOLEAN".


lovasoa:
We should probably also remove at least one of (number, integer, real).

pchemguy:
Here you go

lovasoa:
Thank you!

PR #428, opened 2024-06-19
Corrected SQL query

pchemguy:
It is dangerous to rely on a particular order of a result set, which is not explicitly sorted.


lovasoa:
You are very right. But this query is used all over the website, so I'd rather keep it short. 

We can create a view that will contain only the shell we need and then use that view.

pchemguy:
OK

pchemguy:
Cancelling this change

PR #432, opened 2024-06-20
Parse JSON menu_item for SQLite and skip blank items

pchemguy:
Use parse_json helper to enable the use of the object form of "shell" with JSON containing menu items. Also in this pull request is an explicit filter on blank items (for example, generated via `NULL AS menu_item` or via `'' AS menu_item`). The latter feature simplifier creation of dynamic menus based on certain conditions, such as whether the user is authenticated or not. 

Note, however, `'' AS menu_item` presently fails due to parse_json and should not be used. `'{}' AS menu-item` apparently works fine.

### DIFF
--- a/sqlpage/templates/shell.handlebars
+++ b/sqlpage/templates/shell.handlebars
@@ -76,12 +76,14 @@
                         {{#if (eq (typeof (parse_json this)) 'object')}}
                             {{#with (parse_json this)}}
                                 <li class="nav-item{{#if this.submenu}} dropdown{{/if}}">
-                                <a class="nav-link {{#if this.submenu}}dropdown-toggle{{/if}}" href="{{#if this.link}}{{this.link}}{{else}}#{{/if}}"
-                                    {{#if this.submenu}}data-bs-toggle="dropdown" data-bs-auto-close="outside"{{/if}}
-                                    role="button"
-                                >
-                                    {{this.title}}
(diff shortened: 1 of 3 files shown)


sqlpage/templates/shell.handlebars
```diff
@@ -73,28 +73,36 @@
...
-                                    {{#each this.submenu}}
-                                        <a class="dropdown-item" href="{{this.link}}">
-                                            {{this.title}}
-                                        </a>
-                                    {{/each}}
-                                </div>
-                            {{/if}}
-                            </li>
+                        {{#if (eq (typeof (parse_json this)) 'object')}}
+                            {{#with (parse_json this)}}
```
lovasoa, line 77: This parses the same json string twice

lovasoa (changes requested):

pchemguy:
There was another issue. `'Help' AS menu_item` would be passed to parse_json causing an unhandled error. So I had to create a more elaborate check removing the first parse_json and only keeping it in the "with" block where a preliminary check is passed.

lovasoa:
This is super useful ! 

Can we add an example to the documentation for these "conditional" menu items ? Otherwise people would never know the possibility even exists.

pchemguy:
I have added a basic example. A more elaborate example will be a part of separate pull request.

PR #436, opened 2024-06-20
Adds support for image-file-based button icons

pchemguy:


examples/official-site/sqlpage/migrations/18_button.sql
```diff
@@ -20,6 +20,7 @@ INSERT INTO parameter(component, name, description, type, top_level, optional) S
     ('space_after', 'Whether there should be extra space to the right of the button. In a line of buttons, this will put the buttons before this one on the left, and the ones after on the right.', 'BOOLEAN', FALSE, TRUE),
     ('icon_after', 'Name of an icon to display after the text in the button', 'ICON', FALSE, TRUE),
     ('icon', 'Name of an icon to be displayed on the left side of the button.', 'ICON', FALSE, TRUE),
+    ('img', 'Path to image file (relative. relative to web root or URL) to be displayed on the button.', 'TEXT', FALSE, TRUE),
```
lovasoa, line 23: Cool ! Can we add an example below in the documentation ?

pchemguy: Sure, just did. It relies on the updated component, so to display it correctly, the updated "button" component needs to be integrated into the SQLpage copy serving the official site or you could treat it as a custom overwrite until the server is updated.

lovasoa:
Merci !

lovasoa:
I'd like to rename it to 'image' for consistency with other similar attributes, if that's ok

pchemguy:
Yes, you are right. Updated it

PR #440, opened 2024-06-21
Adds fixed top menu option to shell

pchemguy:
If "fixed_top_menu" boolean is set to True on the "shell" component, the top navigation bar will be fixed at top. However, it spans the entire page width (probably, by design) regardless of layout used and works best with the fluid layout.


lovasoa:
Great, thanks for your contribution !

lovasoa:
@pchemguy I think this introduced a margin issue in some layouts:

image


Can you have a look at it ?

PR #443, opened 2024-06-21
Trims extra white-spaces on the right in menus

pchemguy:


lovasoa (changes requested):
Hello !
  I understand what you want here, but I'm not sure no min width in menus at all is the best option. This minimal menu width is not there by mistake; it's part of https://github.com/tabler/tabler
  Maybe the tabler repository itself would be the right place to discuss this kind of questions. @codecalm has a lot more design experience than me, and I'd like to keep the defaults from tabler when they make sense.

And as a side note, if we really wanted to override default menu width, we would do it in `sqlpage.css`, not inline in `shell.handlebars`

PR #448, opened 2024-06-23
Fixes OR helper usage

pchemguy:
Apparently, the OR helper only supports two arguments, so three arguments require two helper calls

### DIFF
--- a/sqlpage/templates/shell.handlebars
+++ b/sqlpage/templates/shell.handlebars
@@ -87,7 +87,7 @@
                     {{~#each (to_array menu_item)~}}
                         {{~#if (or (eq (typeof this) 'object') (and (eq (typeof this) 'string') (starts_with this '{')))}}
                             {{~#with (parse_json this)}}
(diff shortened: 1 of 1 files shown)


sqlpage/templates/shell.handlebars
```diff
@@ -87,7 +87,7 @@
                     {{~#each (to_array menu_item)~}}
                         {{~#if (or (eq (typeof this) 'object') (and (eq (typeof this) 'string') (starts_with this '{')))}}
                             {{~#with (parse_json this)}}
-                                {{#if (or (gt (len this.title) 0) (gt (len this.icon) 0) (gt (len this.image) 0))}}
+                                {{#if (or (or (gt (len this.title) 0) (gt (len this.icon) 0)) (gt (len this.image) 0))}}
```
lovasoa, line 90: wouldn't this work ?

```suggestion
                                {{#if (or (or this.title this.icon) this.image))}}
```

pchemguy: I had some troubles filtering empty strings with plain checks. I double checked just now and it looks like it does work after all.

lovasoa (approved):

Issue #435, opened 2024-06-20
Spinner to indicate page is loading


djyotta:
### What are you building with SQLPage ?

Personal file, text sharing.

### What is your problem ? A description of the problem, not the solution you are proposing.

I have (rare) situation where some sqlpage.fetch calls are slow (~10 seconds).

In other frameworks, I usually emmit a tag with a spinner container class on it before long running compute, then emmit another tag with a terminating class on it after compute is finished.

I apply the spinner via css. It's display is set to none when the sibling tag with terminating class is emmited.

I tried to do this with the text component, but it seems the partial page is not sent to the browser so it doesn't work.

### What are you currently doing ? Since your solution is not implemented in SQLPage currently, what are you doing instead ?

I have no workaround. Page hangs and it blank.

**Describe the solution you'd like**

I'd like a spinner component that can be selected befor long compute. It could be terminated once next component is selected or by manually selecting a spinner-stop component.

It doesn't have to be a pure css spinner as I described, just whatever works even if it involves JS

It would be ideal if the spinner could be customized by supplying css and/or a class to apply the spinner to.

**Describe alternatives you've considered**

I'm not aware of any alternatives.

Maybe I can do a custom handlebars component?

lovasoa:
Hi ! 

  That would indeed be a nice feature, but probably not easy to implement. What you can do in the meantime is have your long running task run in the background, and update the database on its own once it is done. The `sqlpage.exec` script would just launch the task in the background and return immediately. 
  The SQLPage UI can then display the status of the computation based on the data in the database.

djyotta:
@lovasoa would the user have to refresh the page manually then? I suppose I could use the meta tag to auto refresh - but apparently this is frowned upon...
```
<meta http-equiv="refresh" content="5">
```

lovasoa:
There is already a refresh attribute in the shell component

djyotta:
Is the page buffered in memory server side? It seems no matter whad I do, I can't get partial page sent to browser.

lovasoa:
No, it's not buffered on the server, and each row is rendered and sent to the browser individually. However, browsers have their own buffering and rendering logic.

djyotta:
Ok, this put me on the right track. From what I can tell, while sqlpage isn't buffering server side, it appears that sqlpage.fetch is blocking the browsers asynchronous requests for static content linked in the header (css and js etc). When I delay the sqlpage.fetch until after the browser has loaded the static content, then my spinner shows fine.

[...]

PR #449, opened 2024-06-23
#435: show spinner indicating page is loading

djyotta:


djyotta:
@lovasoa can you take a look to see if it's ok in principle before I start on documentation?

sqlpage/sqlpage.css
```diff
@@ -50,4 +50,46 @@ td > p {
...
+  display: block;
+}
+.sqlpage-spinner-start:not(:last-child) {
+  /* firefox doesn't support :has
+     so hide the spinner if is no longer the last element */
+  display: none;
+}
+
+/* progress indicator container */
+div.sqlpage-spinner-progress {
```
djyotta, line 78: I'm refactoring to use the tabler progress element

examples/show-progress/go.sql
```diff
@@ -0,0 +1,13 @@
+-- can disable the spinner to show only progress bar
+SELECT 'loader-start' AS component,  'non-existent-class' AS spinner;
```
djyotta, line 2: This is a bit janky. I'm thinking instead of default spinner, user must specify a tabler spinner. If NULL then no spinner.

examples/official-site/sqlpage/sqlpage.yaml
```diff
@@ -2,4 +2,7 @@
 database_url: "sqlite::memory:"
 
 # We have a file upload example, and would like to limit the size of the uploaded files 
-max_uploaded_file_size: 256000
\ No newline at end of file
+max_uploaded_file_size: 256000
+
+# For the show-progress example to work, we have to disable compression
+compress_responses: False
```
djyotta, line 8: @lovasoa I'm not sure if this is acceptable for your website. The example works best live (I tried but failed to achieve the inline example style as for the 'form' component).

djyotta: nvm - I forgot I can disable compression with the header

examples/official-site/sqlpage/migrations/01_documentation.sql
```diff
@@ -757,6 +757,69 @@ and `shell.json` would be placed at the website''s root and contain the followin
...
 
+INSERT INTO component(name, icon, description, introduced_in_version) VALUES
+    ('loader-start', 'refresh', 'Display a spinner to indicate page is loading.', '0.25.0');
+
+INSERT INTO parameter(component, name, description_md, type, top_level, optional) SELECT 'loader-start', * FROM (VALUES
+    ('spinner', '
+The name of a [spinner](https://tabler.io/docs/components/spinners) (from tabler.io).
+Default is "spinner-border".
+Set to the empty string to disable the spinner - e.g. to display only progress
+updates.
```
djyotta, line 768: I've noticed this whole approach doesn't work inside run_sql for some reason. So I should note that 'loader-start' component should be selected before descending into run_sql.

lovasoa: Yes, run_sql is a function, it returns a single json value all at once, it can't stream its results.

djyotta: I thought as much but didn't want to assume. It's good to have the clarification.

PR #451, opened 2024-06-24
Daniel/450 readonly disabled for textarea

### LINKED ISSUE #450: readonly and disabled are not applied to text area form elements
### Introduction

The textarea input field can not be disabled or readonly.

### To Reproduce

List of steps to reproduce the behavior. Include the sql file you are using and the eventual relevant parts of your database schema

```sql
SELECT 'form' AS 'component';
SELECT 'Paste something here...' AS placeholder
, 'content' AS name
, 'textarea' AS type
, '' AS label
, TRUE AS disabled
;
```

Field should not be editable, but it is.

### Version information

 - OS: MXLinux
 - Database: N/A
 - SQLPage Version: 0.24.0

djyotta:
Closes #450 

Tested with this:
```sql
SELECT 'form' AS component, 'post' AS method;
SELECT 'select' AS type
, json('[
  { "label": "a", "value": "a", "selected": true },
  { "label": "b", "value": "b", "selected": false },
  { "label": "c", "value": "b", "selected": true },
]') AS options
, TRUE AS multiple
, TRUE AS dropdown
, TRUE AS readonly
, 's' AS name
;
SELECT 'textarea' AS type
, 'abc' AS value
, 'ta' AS name
, TRUE AS readonly
;
```
* disabled works fine for textarea and select and dropdown select
* readonly works fine for textarea but not for dropdown select, so I removed implementation of readonly on select altogether

### DIFF
--- a/sqlpage/templates/form.handlebars
+++ b/sqlpage/templates/form.handlebars
@@ -64,6 +64,8 @@
                         {{~#if id}} id="{{id}}" {{/if~}}
                         {{~#if required}} required="required" {{/if~}}
                         {{~#if autofocus}} autofocus {{/if~}}
+                        {{~#if disabled}}disabled {{/if~}}
+                        {{~#if readonly}}readonly {{/if~}}
                         {{~#if multiple}} multiple {{/if~}}
                         {{~#if (or dropdown searchable)}} 
                             data-pre-init="select-dropdown"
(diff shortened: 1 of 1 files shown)


lovasoa (approved):
Thank you for your contribution !

djyotta:
No problem. I really needed this feature.

PR #463, opened 2024-06-25
Do not ignore points with a latitude of 0 in map.handlebars

matthieupeeters:
The points that are shown on the map are filtered on {{~if latitude}}. By adding "inculdeZero=true" the points that have a latitude of 0 will also be shown.


lovasoa (approved):
Thank you for your contribution!

PR #469, opened 2024-06-27
Fixes top bar being shown only when page title is defined

pchemguy:
When any top bar element is defined (title, icon, image, or menu), show top bar.


sqlpage/templates/shell.handlebars
```diff
@@ -63,7 +63,7 @@
 
 <body class="layout-{{default layout 'boxed'}}" {{#if theme}}data-bs-theme="{{theme}}"{{/if}}>
 <div class="page">
-    {{#if title}}
+    {{#if (or (or title (or icon image)) (gt (len (to_array menu_item)) 0))}}
```
lovasoa, line 66: This converts menu_item to an array and then immediately throws the array away. 

Can we say that we show the bar if menu_item is defined, whatever its contents?

lovasoa (changes requested):

sqlpage/templates/shell.handlebars
pchemguy: Right. I checked its type directly and it returned null (I do not know why), which is why I had to switch to this construct. May be there is a better way.

lovasoa: yes variables that are not defined are observed as null

pchemguy: Right, those NOT defined. Why it returns null for menu_item (meaning the test for null, like with icon and image does not work) is unclear. I mean, I of course tested both variants and switched to this logically unnecessary construct because the direct test didn't work. Like `typeof menu_item` = null AND the length of converted array was 7 in my test case.

lovasoa:
Thank you @pchemguy !

PR #486, opened 2024-07-06
(feature) add sqlpage.query_string()

guspower:
Adds the `sqlpage.query_string()` SQLPage function.

There are a few auth-related scenarios where it is useful to be able to directly access the query string and propagate it, for example when redirecting to a login page from a url with query string, and then redirecting the user back to that url on successful login.

There may be better ways to do this, but here is a simple `query_string` function for your consideration.

### DIFF
--- a/src/webserver/database/sqlpage_functions/functions.rs
+++ b/src/webserver/database/sqlpage_functions/functions.rs
@@ -21,6 +21,7 @@ super::function_definition_macro::sqlpage_functions! {
     header((&RequestInfo), name: Cow<str>);
 
     path((&RequestInfo));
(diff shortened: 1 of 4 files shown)


guspower:
Yep that seems reasonable - in fact it was what I started out doing but then realized that I'd have to take a perspective on both ordering and encoding. Let me rework this later in the week and come back to you with something that fits.

lovasoa:
@guspower , what do you think about doing this instead: https://github.com/lovasoa/SQLpage/pull/494 ?

guspower:
Yep that looks good to me. Much cleaner +1

PR #523, opened 2024-08-10
Merge & correct the layout of the sidebar

Ajotah98:
I changed the layout to correct the grid problem on the header.

Also, I wasn't aware that the sidebar view also had the same problem for long titles, so I added a display inline that makes a better UI (in my opinion)

### DIFF
--- a/sqlpage/templates/shell.handlebars
+++ b/sqlpage/templates/shell.handlebars
@@ -41,7 +41,7 @@
         </style>
         {{/if}}
     {{/if}}
-    
+
     <script src="{{static_path 'sqlpage.js'}}" defer nonce="{{@csp_nonce}}"></script>
     {{#each (to_array javascript)}}
         {{#if this}}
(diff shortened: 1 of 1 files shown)


lovasoa:
@Ajotah98 , I think you have a formatter that changed a bunch of files that are not related to the PR. Can you revert the formatting changes and keep only the actual shell component updates?

sqlpage/templates/shell.handlebars
```diff
@@ -147,14 +147,14 @@
...
-            <header id="sqlpage_header"></header>
+            <header id="sqlpage_header">
         {{#if sidebar}}
         <aside class="navbar navbar-vertical navbar-expand-lg" {{#if sidebar-theme}}data-bs-theme="dark" {{/if}}>
             <div class="container-fluid">
                 <button class="navbar-toggler collapsed" type="button" data-bs-target="#sidebar-menu" aria-controls="sidebar-menu" data-bs-toggle="collapse" aria-expanded="false" aria-label="Toggle navigation">
                     <span class="navbar-toggler-icon"></span>
                 </button>
-                <h1 class="navbar-brand navbar-brand-autodark">
+                <h1 class="navbar-brand navbar-brand-autodark" style="display:inline;">
```
lovasoa, line 157: can we use a bootstrap class instead of an inline style ?

lovasoa (changes requested):

lovasoa:
additionnally, this seems to break long titles ellipsis

image

lovasoa:
Ok, I think everything is fixed now

Ajotah98:
I apologize!! I couldn't see this until today. 

I'm grateful you managed to repair the broken things, and so sorry for breaking the title.
Thanks!!

PR #543, opened 2024-08-18
Disable CA host certificat on test.

guiguid:
This patch disable the use of system CA root.
fix https://github.com/lovasoa/SQLpage/issues/541


lovasoa (approved):
Thank you, @guiguid !

PR #544, opened 2024-08-18
feat: add 404.sql handler

### LINKED ISSUE #529: allow catch all `.sql` files for custom routing
What are you building with SQLPage ?

I'd like to build a [pixiecore api](https://github.com/danderson/netboot/blob/main/pixiecore/README.api.md#api-specification) server based on SQL Page.

What is your problem ? A description of the problem, not the solution you are proposing.

The server has to answer requests to the following path: `<apiserver-prefix>/v1/boot/<mac-addr>`.
The last part of the path is the actual argument to the API call. To the best of my knowledge, there is no way in SQLpage to have a one or multiple arbitrary paths served by a single `.sql` file. There is of course `index.sql`, but that maps one path to one `.sql` file.

What are you currently doing ? Since your solution is not implemented in SQLPage currently, what are you doing instead ?

Introducing a map in the configuration, mapping URL paths to `.sql` files. When the current URL path starts with the key of an entry from that map, use the corresponding value from the map as path to a `.sql` file and execute it. Otherwise, carry on as usual.

**Describe alternatives you've considered**

Having a special name, e.g. if in a directory there is a `_catch_all.sql` then only that is considered for any URL whose path goes into this directory.

wucke13:
Fixes #529 .

@lovasoa Is this an acceptable business logic to you? I'm sure the code can be cleaned up, and there might be an argument for using `PathBuf` directly instead of doing `String` manipulation, but besides that?

### DIFF
--- a/src/webserver/http.rs
+++ b/src/webserver/http.rs
@@ -374,9 +374,9 @@ fn path_to_sql_file(path: &str) -> Option<PathBuf> {
 }
 
 async fn process_sql_request(
-    mut req: ServiceRequest,
+    req: &mut ServiceRequest,
     sql_path: PathBuf,
-) -> actix_web::Result<ServiceResponse> {
+) -> actix_web::Result<HttpResponse> {
     let app_state: &web::Data<AppState> = req.app_data().expect("app_state");
     let sql_file = app_state
         .sql_file_cache
@@ -384,8 +384,7 @@ async fn process_sql_request(
         .await
         .with_context(|| format!("Unable to get SQL file {sql_path:?}"))
         .map_err(anyhow_err_to_actix)?;
-    let response = render_sql(&mut req, sql_file).await?;
-    Ok(req.into_response(response))
+    render_sql(req, sql_file).await
 }
 
 fn anyhow_err_to_actix(e: anyhow::Error) -> actix_web::Error {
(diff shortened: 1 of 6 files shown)


lovasoa:
Hi @wucke13 ! The logic seems to match exactly what we discussed; everything good for me! 

If you can clean up the code, and maybe put it in a separate function, I'll merge and this will be in the next version. 

Thank you for your contribution!

lovasoa:
I'm not sure what the best way to teach this new feature to users is. I think we should have 404.sql files in at least some of the examples on this repository. 

We could also mention its existence in the documentation of the json component and the path function. We could also mention it in configuration.md.

wucke13:
I definitely plan on adding an example. I'm quite indifferent on one thing: Should we do the path traversal stuff using a `String` or a `PathBuf`. What I currently like about using a `String`, is that we can discern `a/b/c/` and `/a/b/c`. In the former case we want to check `/a/b/c/404.sql`, but in the latter case we don't. I think this nuance gets lost when converting the `String` to a `PathBuf`. @lovasoa Would you agree on using a `String`?

lovasoa:
Yes, we should use a string and not a `PathBuf`. If there are other places where I used a PathBuf to represent a URL path, it is a mistake. `PathBuf`s represent a platform-specific local file path; we want URL path handling to work exactly the same on all platforms.

wucke13:
Ok, I now have a cleaner solution. Please let me know if you are happy with it as well. If yes, I'll also push an example, squash the first two commits and hopefully then we are ready for merge :smile: .

src/webserver/http.rs
```diff
@@ -450,56 +497,15 @@ pub async fn main_handler(
...
-                        log::trace!("The 404 handler {sql_path:?} does not exist");
-                        full_path.truncate(idx);
-                        continue;
-                    };
-
-                    log::debug!("Processing SQL request via fallback: {:?}", sql_path);
-                    return process_sql_request(service_request, sql_path).await;
-                }
-            }
+            Err(e) => serve_fallback(service_request, e).await,
```
lovasoa: I think we should call the 404 handler only if the original error is a 404; not for any error

src/webserver/http.rs
```diff
@@ -432,6 +432,53 @@ async fn serve_file(
...
+        .chain(std::iter::once(0))
+    {
+        // Remove the trailing substring behind the current `/`, and append `404.sql`.
+        maybe_fallback_path.truncate(idx);
+        maybe_fallback_path.push_str(&catch_all);
+
+        // Check if `maybe_fallback_path` actually exists, if not, skip to the next round (which
+        // will check `maybe_fallback_path`s parent directory for fallback handler).
+        let sql_path = PathBuf::from(&maybe_fallback_path);
+        let app_state: &web::Data<AppState> = service_request.app_data().expect("app_state");
```
lovasoa, line 462: we should do that just once, out of the loop

src/webserver/http.rs
```diff
@@ -432,6 +432,53 @@ async fn serve_file(
...
+        if let Err(_) = app_state
+            .sql_file_cache
+            .get_with_privilege(app_state, &sql_path, false)
+            .await
+        {
+            log::trace!("The 404 handler {sql_path:?} does not exist");
+            continue;
```
lovasoa: we should continue (and ignore the error) only if the error is indeed a file not found. For any other error, we want to return the error

lovasoa (changes requested):
Thank you, the overall logic looks good to me !

But we should be careful to invoke the 404 handler only for actual 404s, not for other errors. Maybe we can add tests for that ?

src/webserver/http.rs
wucke13: Done

src/webserver/http.rs
wucke13: Done.

src/webserver/http.rs
wucke13: Done.

wucke13:
@lovasoa I adjusted again. Most notably, now the `serve_fallback` call is only made when the error is `404`:

````rust
// On 404/NOT_FOUND error, fall back to `404.sql` handler if it exists
let response = match maybe_response {
    // File could not be served due to a 404 error. Try to find a user provide 404 handler in
    // the form of a `404.sql` in the current directory. If there is none, look in the parent
    // directeory, and its parent directory, ...
    Err(e) if e.as_response_error().status_code() == StatusCode::NOT_FOUND => {
        serve_fallback(&mut service_request, e).await?
    }

    // Either a valid response, or an unrelated error that shall be bubbled up.
    e => e?,
};
```

I'm not sure how to test this elegantly, though at least at a superficial glance this seems to be a pretty robust implementation of `only ever try to do something when there is an `404 Not Found` error.

I also added an example, that both illustrates the use and explains the fallback mechanism in prose. While I did mention custom error handling and implementing REST APIs as potential use-cases, I intentionally did not pitch the use of this feature as gateway to dynamic routing in general, I feel that goes against the overall spirit of SQLpage. Creative users anyway will do what they want :smile: , but we don't have to encourage it in the examples.

lovasoa:
Great! Thank you @wucke13 ! 

You can add a test in tests/index.rs that makes a request to tests/xyz.sql that doesn't exist, and add a tests/404.sql. 

Maybe we can still mention using it for APIs in the documentation of the json component.

wucke13:
@lovasoa  Done. Also squashed a little for a clean history.

src/webserver/http.rs
```diff
@@ -434,25 +433,89 @@ async fn serve_file(
...
+        fallback_path_candidate.push_str(&catch_all);
+
+        // Check if `maybe_fallback_path` actually exists, if not, skip to the next round (which
+        // will check `maybe_fallback_path`s parent directory for fallback handler).
+        let mabye_sql_path = PathBuf::from(&fallback_path_candidate);
+        let Ok(sql_file) = app_state
+            .sql_file_cache
+            .get_with_privilege(app_state, &mabye_sql_path, false)
+            .await
+        else {
```
lovasoa, line 470: We have to check if the error here is indeed a file not found. If it's another error, then we can stop end return the error.

src/webserver/http.rs
```diff
@@ -434,25 +433,89 @@ async fn serve_file(
...
+
+    let req_path = req_path(&service_request);
+    let mut fallback_path_candidate = req_path.clone().into_owned();
+    log::debug!("Trying to find a {catch_all:?} handler for {fallback_path_candidate:?}");
+
+    let app_state: &web::Data<AppState> = service_request.app_data().expect("app_state");
+
+    // Find the indeces of each char which follows a directroy separator (`/`). Also consider the 0
+    // index, as we also have to try to check the empty path, for a root dir `404.sql`.
+    for idx in req_path
```
lovasoa, line 454: Thinking about security here: could an attacker dos us by sending a request that is just thousands of slashes? Maybe we should add a (large) limit to the number of iterations?

tests/index.rs
```diff
@@ -51,6 +51,19 @@ async fn test_404() {
...
+#[actix_web::test]
+async fn test_404_fallback() {
+    for f in [
+        "/tests/does_not_exist.sql",
+        "/tests/does_not_exist.html",
+        "/tests/does_not_exist/",
+    ] {
+        let resp_result = req_path(f).await;
+        let resp = resp_result.unwrap();
+        assert_eq!(resp.status(), http::StatusCode::OK, "{f} isnt 200");
```
lovasoa, line 63: Can we also assert that the contents match the sentence from the 404.sql file!

lovasoa (changes requested):
We are almost there!

src/webserver/http.rs
wucke13: Whats special about this condition? The same issue arises when a passing a thousand get parameters, except that the parsing is more expensive for GET parameters, while for the fallback the file syscalls are more expensive. I get the idea, but I'm against starting to do this here. Either there is a generic way that is implemented in the actix-web to do this kind of DDoS protection, or there isn't, but I don't think it is clever to start limiting at arbitrary places by hand. Also, I'd assume that there already is a max header size, which naturally will limit the amount of iterations done here anyway.

src/webserver/http.rs
wucke13: This one does not spark joy, grr, `anyhow`.

wucke13: Done.

wucke13: Done for real.

tests/index.rs
wucke13: Yes!

wucke13:
I'd be glad if we can get this done rather sooner or later, my motivation is somewhat depleted.

src/webserver/http.rs
lovasoa: Yes, indeed, I was thinking about doing thousands of syscalls, or more importantly, database queries since the files can be stored in the database. I'll add the limit.

lovasoa (approved):

lovasoa:
Thank you @wucke13 ! This PR is going to be really useful! I'll try to reference the new feature more in the docs to make it more discoverable

src/webserver/http.rs
wucke13: If you add a limit, make it configurable please

PR #550, opened 2024-08-21
Add new component Accordion

Ajotah98:
I create the PR to add the new Accordion component based on what we discussed on #547 

Let me know if is needed something more in the documentation file if necessary


lovasoa:
can name the parameters title, description, and description_md, so that it is compatible with the list, map, card, and datagrid component ? I think it reduces the mental overhead, and it makes it easy to test different components to display the same data.

lovasoa:
Thanks @Ajotah98 ! I made a few changes in https://github.com/lovasoa/SQLpage/commit/38a70a14063264ad0fbbdad5bd7e9f0a556c4997

PR #615, opened 2024-09-22
feat: add lazy component

mtt-artis:
Hi 👋,

I'm working with some resource-intensive queries that take some time to execute, and the `embed` option on the `card` component works really well. However, I’d like to prevent layout shifts and avoid having cards nested inside other cards.

The `lazy` component follows a similar logic but provides an element you can style before it's replaced with the actual content.

Please feel free to make any changes, rename the component (coming from React/Solid, that's what comes naturally to me), or guide me in the right direction.

If you like this implementation, I'd be happy to add it to the documentation.

### DIFF
--- a/examples/cards-with-remote-content/lazy.sql
+++ b/examples/cards-with-remote-content/lazy.sql
@@ -1,0 +1,10 @@
+select 'lazy' as component;
+
+select 
+    '/chart-example.sql?_sqlpage_embed' as embed,
+    'card my-2' as class,
+    'height:340px' as style;
+
+select '/map-example.sql' as embed;
+
+select '/table-example.sql?_sqlpage_embed' as embed;
--- a/sqlpage/sqlpage.js
+++ b/sqlpage/sqlpage.js
@@ -3,47 +3,50 @@
 
 const nonce = document.currentScript.nonce;
 
-function sqlpage_card() {
-    for (const c of document.querySelectorAll("[data-pre-init=card]")) {
-        const source = c.dataset.embed;
-        fetch(c.dataset.embed)
-            .then(res => res.text())
-            .then(html => {
-                const body = c.querySelector(".card-content");
-                body.innerHTML = html;
-                c.removeAttribute("data-pre-init");
-                const spinner = c.querySelector(".card-loading-placeholder");
-                if (spinner) {
-                    spinner.parentNode.removeChild(spinner);
-                }
-                const fragLoadedEvt = new CustomEvent("fragment-loaded", {
-                    bubbles: true
-                });
-                c.dispatchEvent(fragLoadedEvt);
-            })
+function sqlpage_embed() {
+  for (const c of document.querySelectorAll("[data-embed]")) {
+    if (c.ariaBusy === "true") return;
+    c.ariaBusy = true;
+    const [source, params] = c.dataset.embed.split("?");
+    if (!source) return;
+    const search = new URLSearchParams(params);
+    if (!search.has("_sqlpage_embed")) {
+      search.set("_sqlpage_embed", "true");
     }
+    fetch(`${source}?${search}`)
+      .then(res => res.text())
+      .then(html => {
+        c.innerHTML = html;
+        c.ariaBusy = false;
+        delete c.dataset.embed;
+        c.dispatchEvent(new CustomEvent("fragment-loaded", {
+          bubbles: true
+        }));
+      })
+      .catch(err => console.error("Fetch error: ", err));
+  }
 }
 
-function sqlpage_table(){
-    // Tables
-    for (const r of document.querySelectorAll("[data-pre-init=table]")) {
-        new List(r, {
-            valueNames: [...r.getElementsByTagName("th")].map(t => t.textContent),
-            searchDelay: 100,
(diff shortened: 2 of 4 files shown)


lovasoa:
Hello ! This sounds like a very useful component to add, and I'll be happy to integrate it. Thank you !

Could you please:
 - remove the indentation changes to make the code easier to review
 - add documentation for the component in https://github.com/lovasoa/SQLpage/tree/main/examples/official-site/sqlpage/migrations
   - The audience of SQLPage is generally not familiar with web development, so we need to find a way to explain the concept and why it is useful in a way that is approachable to non-technical users.
 - add a test for it in https://github.com/lovasoa/SQLpage/blob/main/tests/end-to-end/official-site.spec.ts

sqlpage/sqlpage.js
```diff
@@ -3,47 +3,50 @@
...
-                if (spinner) {
-                    spinner.parentNode.removeChild(spinner);
-                }
-                const fragLoadedEvt = new CustomEvent("fragment-loaded", {
-                    bubbles: true
-                });
-                c.dispatchEvent(fragLoadedEvt);
-            })
+function sqlpage_embed() {
+  for (const c of document.querySelectorAll("[data-embed]")) {
```
lovasoa, line 7: Maybe this should be done in parallel rather than sequentially ?

lovasoa (changes requested):

sqlpage/sqlpage.js
```diff
@@ -3,47 +3,50 @@
...
+    if (c.ariaBusy === "true") return;
+    c.ariaBusy = true;
+    const [source, params] = c.dataset.embed.split("?");
+    if (!source) return;
```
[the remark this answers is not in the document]
mtt-artis: my bad, rest of a forEach variant

sqlpage/sqlpage.js
mtt-artis: I am not sure if I understand you.
We don't await the fetch so the loop does't wait the promise to be resolved to start the next element.

mtt-artis:
hi 👋, 
Thanks for your comments

> - remove the indentation changes to make the code easier to review

could we invest for future change and make the indentation consistent across the file ?

I can add tests and documentation during the next WE if you don't mind 😉

examples/cards-with-remote-content/lazy.sql
```diff
@@ -0,0 +1,10 @@
+select 'lazy' as component;
+
+select 
+    '/chart-example.sql?_sqlpage_embed' as embed,
```
mtt-artis, line 4: should we keep embed or lazy here ?

lovasoa:
Yes, we can set an indentation standard, but maybe after your PR? You are going to end up with large conflicts for nothing, otherwise. I need a clean diff in this PR to review it.

examples/cards-with-remote-content/lazy.sql
lovasoa: I think `as embed` is good

sqlpage/templates/lazy.handlebars
```diff
@@ -0,0 +1,13 @@
+<div class="{{default class "my-2"}}"{{~#if style}} style="{{style}}" {{/if~}}>
```
lovasoa, line 1: In all other components, `class` adds new classes without replacing the existing ones

sqlpage/sqlpage.js
```diff
@@ -3,24 +3,30 @@
...
-            })
+function sqlpage_embed() {
+  for (const c of document.querySelectorAll("[data-embed]")) {
+    if (c.ariaBusy === "true") continue;
+    c.ariaBusy = true;
+    let url;
+    try {
+      url = new URL(c.dataset.embed, window.location.href)
+    } catch {
+      console.erreur(`'${c.dataset.embed}' is not a valid url`)
```
lovasoa, line 14:
```suggestion
      console.error(`'${c.dataset.embed}' is not a valid url`)
```

sqlpage/sqlpage.js
```diff
@@ -3,24 +3,30 @@
...
-                if (spinner) {
-                    spinner.parentNode.removeChild(spinner);
-                }
-                const fragLoadedEvt = new CustomEvent("fragment-loaded", {
-                    bubbles: true
-                });
-                c.dispatchEvent(fragLoadedEvt);
-            })
+function sqlpage_embed() {
+  for (const c of document.querySelectorAll("[data-embed]")) {
```
mtt-artis, line 7: is the query better ?
`document.querySelectorAll("[data-embed]:not([aria-busy=true])")`

sqlpage/templates/lazy.handlebars
mtt-artis: i know and i hesitate, this div is a light wrapper and i think user should be able to reset it easily if needed.

same for `<div class="{{default class "card my-2"}}">`. i add this default because most of the component are card.

should we just give no default and handle it like the style param ?

sqlpage/sqlpage.js
lovasoa: Yes it's better !

lovasoa:
Hello ! Any news on this ? I think only docs and tests are missing, then we can merge !

PR #643, opened 2024-10-10
Pull request to add robots.txt and sitemap.xml files (v1)

alexisrc1:


examples/official-site/robots.txt
```diff
@@ -0,0 +1,8 @@
+User-agent: *
+Disallow: /examples/authentification/  # Block the folder "authentification" to avoid blocking by password
+Disallow: /examples/hash_password.sql  # Block access to specific sensitive SQL files like this
+Disallow: /Dockerfile  # Block access to Dockerfile
```
lovasoa, line 4: Why ? https://sql.datapage.app/Dockerfile

examples/official-site/robots.txt
```diff
@@ -0,0 +1,8 @@
+User-agent: *
+Disallow: /examples/authentification/  # Block the folder "authentification" to avoid blocking by password
+Disallow: /examples/hash_password.sql  # Block access to specific sensitive SQL files like this
```
lovasoa, line 3: Why ? This is a useful page, it is not password protected: https://sql.datapage.app/examples/hash_password.sql

examples/official-site/robots.txt
```diff
@@ -0,0 +1,8 @@
+User-agent: *
+Disallow: /examples/authentification/  # Block the folder "authentification" to avoid blocking by password
```
lovasoa, line 2: Maybe do not block the entire authentication example ? We still want it referenced. Ideally change the auth example so the password protected page has a different url, that you can block

examples/official-site/sitemap.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+  <url>
+    <loc>https://sql.datapage.app/</loc>
+    <lastmod>2024-10-10</lastmod>
```
lovasoa, line 5: Is it ok to hardcode the date like this ? Won't the prevent re-crawling when we update the page ?

examples/official-site/sitemap.xml
```diff
@@ -0,0 +1,23 @@
...
+  <url>
+    <loc>https://sql.datapage.app/documentation.sql</loc>
+    <lastmod>2024-10-10</lastmod>
+    <priority>0.80</priority>
+  </url>
+  <url>
+    <loc>https://sql.datapage.app/your-first-sql-website/</loc>
+    <lastmod>2024-10-10</lastmod>
+    <priority>0.80</priority>
+  </url>
```
lovasoa, line 22: Is it ok to have such a little subset of the site in the sitemap ? Won't it penalize other pages ?

We should probably either not have a sitemap, or have one that is generated dynamically in sql

lovasoa (changes requested):

examples/official-site/robots.txt
alexisrc1: which one is ?

examples/official-site/robots.txt
alexisrc1: I was told to limit the number of files that need to be crawled so I figure this one is not useful for Google

examples/official-site/sitemap.xml
alexisrc1: if it remains hard coded for too long it is an issue, for this first try it is ok

PR #694, opened 2024-11-16
add `empty_option` to the form component

xylle:
Add empty option to select

### DIFF
--- a/examples/official-site/sqlpage/migrations/01_documentation.sql
+++ b/examples/official-site/sqlpage/migrations/01_documentation.sql
@@ -251,7 +251,7 @@ INSERT INTO parameter(component, name, description, type, top_level, optional) S
     ('max', 'The maximum value to accept for an input of type number', 'REAL', FALSE, TRUE),
     ('checked', 'Used only for checkboxes and radio buttons. Indicates whether the checkbox should appear as already checked.', 'BOOLEAN', FALSE, TRUE),
     ('multiple', 'Used only for select elements. Indicates that multiple elements can be selected simultaneously. When using multiple, you should add square brackets after the variable name: ''my_variable[]'' as name', 'BOOLEAN', FALSE, TRUE),
-    ('emptyoption', 'Used only for select elements. Add empty option at first', 'BOOLEAN', FALSE, TRUE),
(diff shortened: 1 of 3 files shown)


lovasoa:
Thank you ! You also need to document the new option, otherwise no one will know about it.

The documentation for the form component is here: https://github.com/sqlpage/SQLPage/blob/main/examples/official-site/sqlpage/migrations/01_documentation.sql#L232-L273

lovasoa:
Ideally, you would also update the select example below, to demonstrate the usage of the new option.

xylle:
Like that, it's good ?

examples/official-site/sqlpage/migrations/01_documentation.sql
```diff
@@ -251,6 +251,7 @@ INSERT INTO parameter(component, name, description, type, top_level, optional) S
     ('max', 'The maximum value to accept for an input of type number', 'REAL', FALSE, TRUE),
     ('checked', 'Used only for checkboxes and radio buttons. Indicates whether the checkbox should appear as already checked.', 'BOOLEAN', FALSE, TRUE),
     ('multiple', 'Used only for select elements. Indicates that multiple elements can be selected simultaneously. When using multiple, you should add square brackets after the variable name: ''my_variable[]'' as name', 'BOOLEAN', FALSE, TRUE),
+    ('emptyoption', 'Used only for select elements. Add empty option at first', 'BOOLEAN', FALSE, TRUE),
```
lovasoa, line 254:
```suggestion
    ('empty_option', 'Only for inputs of type `select`. Adds an empty option before the ones defined in `options`.', 'BOOLEAN', FALSE, TRUE),
```

sqlpage/templates/form.handlebars
```diff
@@ -75,6 +75,7 @@
                         {{~#if placeholder}} placeholder="{{placeholder}}" {{/if~}}
                         {{~#if create_new}} data-create_new={{create_new}} {{/if~}}
                         >
+                        {{~#if emptyoption}}<option value=""></option>{{/if~}}
```
lovasoa, line 78:
```suggestion
                        {{~#if empty_option}}<option value=""></option>{{/if~}}
```

lovasoa (changes requested):
I suggest we name it `empty_option` to follow the snake_case wording of other sqlpage component properties.

Issue #743, opened 2024-12-19
v0.31.0 - Component "card" is displayed incorrectly


Floflobel:
### Introduction

I have a component card that doesn't display very well in the latest version v0.31.0. I have rollback to v0.30.1 for the moment.

### To Reproduce

```sql
select 
    'card'                 as component,
    2                      as columns;
select 
    'Information'         as title,
    'settings'             as icon,
    'green'                as color,
    '/sftp_detail_properties.sql?_sqlpage_embed&name=' || $name as embed;
select 
    'Statistics'           as title,
    'chart-dots-2'         as icon,
    'orange'               as color,
    '/sftp_detail_statistics.sql?_sqlpage_embed&name=' || $name as embed;
```

### Screenshots
v0.31.0:
![Screenshot from 2024-12-19 15-37-14](https://github.com/user-attachments/assets/0a7a786e-c8d6-49ed-a99d-e8c16407ba13)

v0.30.1:
![Screenshot from 2024-12-19 15-37-44](https://github.com/user-attachments/assets/277fd5a7-e169-4c4c-97fb-94a92273f696)

### Version information

 - OS: Debian 12
 - Database [e.g. SQLite, Postgres] Postgres
 - SQLPage Version [found when hovering the default footer of pages]: 0.31.0

lovasoa:
Hi !
 v0.31 removed borders from cards that have embedded contents.
 
  We should probably remove the border around embedded contents only when there is not title or description in the card.
  You can overwrite `sqlpage/templates/card.handlebars` locally if this is blocking you from upgrading to v0.31.

  Do you think you could also open a pull request against https://github.com/sqlpage/SQLPage/blob/main/sqlpage/templates/card.handlebars#L29 ?

Issue #610, opened 2024-09-22
Ability to set 0 margin for embedded cards


accforgithubtest:
> What is your problem ? A description of the problem, not the solution you are proposing.

I seem to be wasting a non-trivial amount of screen real estate due to embedding cards to 2 or 3 levels (i.e a card which embeds a sql file, which in turn is a card that embeds another sql which is also a card ... ).
FYI - I needed to create the 2 or 3 levels of embedded cards because because the video url from go2rtc doesn't behave as expected (height and width issues) when embedded directly into parent page. This is a url for security cam video feed embedded into a sqlpage screen which is displayed in a kiosk mode screen on a 27" 1080p monitor. 

The feature request is about not wasting few pixels of margin created around each card, which multiples with embedded cards. 

> What are you currently doing ? Since your solution is not implemented in SQLPage currently, what are you doing instead ?

If there are any ways to set 0 margin for cards with embedded sql, please let me know.

> A clear and concise description of what you want to happen.

Would like to be able to set `0 as margin` either at the top-level parameter, or at row level parameter which might give more options to control the margins.

> A clear and concise description of any alternative solutions or features you've considered.

If there are any ways to set 0 margin for cards with embedded sql, please let me know.

Thank you for your time and consideration !

lovasoa:
Hello and sorry for the (very) late response.

You are correct that the card margins are wasting space, and confusing users when using the `embed` property to create complex layouts.

I fixed the problem, and the result will be in SQLPage v0.31. Better late than never :)

![image](https://github.com/user-attachments/assets/ea85438d-5fcb-4eed-b90b-a4385675355d)

accforgithubtest:
Thank you so much for this! It does look great now.
Cheers!

PR #744, opened 2024-12-20
fix cards that have embedded contents

Floflobel:
Related to https://github.com/sqlpage/SQLPage/issues/743

### DIFF
--- a/examples/official-site/sqlpage/migrations/31_card_docs_update.sql
+++ b/examples/official-site/sqlpage/migrations/31_card_docs_update.sql
@@ -24,7 +24,7 @@ INSERT INTO parameter(component, name, description, type, top_level, optional) S
     ('footer_link', 'An URL to which the user should be taken when they click on the footer.', 'URL', FALSE, TRUE),
     ('style', 'Inline style property to your iframe embed code. For example "background-color: #FFFFFF"', 'TEXT', FALSE, TRUE),
     ('icon', 'Name of an icon to display on the left side of the card.', 'ICON', FALSE, TRUE),
-    ('color', 'The name of a color, to be displayed on the left of the card to highlight it.', 'COLOR', FALSE, TRUE),
+    ('color', 'The name of a color, to be displayed on the left of the card to highlight it. If the embed parameter is enabled and you don''t have a title or description, this parameter won''t apply.', 'COLOR', FALSE, TRUE),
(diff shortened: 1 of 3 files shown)


lovasoa:
It looks like this PR re-adds the double-border issue that was removed in v0.31 on the graphs from the example page : 


image

with this pr:

image

The problem was initially reported by @accforgithubtest in https://github.com/sqlpage/SQLPage/issues/610


It is clear that we should not disable the borders when we have `color` or `icon` set, but I'm not sure what the best behavior is when there is only a title. What do you think ?

accforgithubtest:
> It is clear that we should not disable the borders when we have color or icon set, but I'm not sure what the best behavior is when there is only a title. What do you think ?

I might be the only one who still thinks the current behavior in v0.31 is better, when there is only a title in there. To me the double border looks worse off somehow. 
Originally I raised the issue because this would multiply the borders and empty spaces when there are 3 levels of embeded cards, etc.

lovasoa:
> If I call my embeded with the “?_sqlpage_embed” parameter, there's a border.

We still need to separate components from one another in the embedded page. I don't think a missing border would look good when there are multiple components in the target page.

> Would be great if we can keep the behavior from the recent fix where each embedded card does add empty space and border when there is no title / text in the card, but just another embed (either iframe or sqlpage_embed).

This pull request keeps the border when there is no title.

---

```sql
select 'card' as component;
select 'xbed.sql?_sqlpage_embed' as embed;
select 'xbed.sql?_sqlpage_embed' as embed, 'red' as background_color;

select 'card' as component;
select 'xbed.sql?_sqlpage_embed' as embed, 'title' as title;
select 'xbed.sql?_sqlpage_embed' as embed, 'green' as color, 'tool' as icon;
```

now gives

image


This is not perfect, but strictly better than what we had before, I think.
This allows both your usecases: cards used strictly for layout, including deeply embedded cards (like @accforgithubtest uses), and decorated cards with just one level of embedding (like @Floflobel uses).

lovasoa:
>     `<fieldset class="form-fieldset mb-1 p-0 {{#if _sqlpage_embed}}border-0 bg-white{{/if}}">`

@Floflobel : this looks nice in your example, but would require special handling of sqlpage_embed in all components, not just the form component...

lovasoa:
I think we can already merge this, since it's a strict improvement over both v0.30 (where you couldn't have multiple layers of embedding) and v0.31 (where embedding with card titles or icons looked ugly).

PR #794, opened 2025-01-27
Handle requests without file extension

lovasoa:
replaces https://github.com/sqlpage/SQLPage/pull/786

### DIFF
--- a/src/webserver/http.rs
+++ b/src/webserver/http.rs
@@ -34,6 +34,8 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::time::SystemTime;
 use tokio::sync::mpsc;
+use crate::webserver::routing::{calculate_route, AppFileStore};
+use crate::webserver::routing::RoutingAction::{NotFound, Execute, CustomNotFound, Redirect, Serve};
 
 #[derive(Clone)]
 pub struct RequestContext {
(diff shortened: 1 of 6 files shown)


src/webserver/routing.rs
```diff
@@ -0,0 +1,496 @@
...
+        result.push_str(
+            uri.path()
+                .strip_prefix(config.prefix().path())
+                .expect("Unable to remove expected prefix from path"),
+        );
+
+        Ok(PathBuf::from(result))
+    } else {
+        Err(Redirect(config.prefix().clone()))
+    }
```
lovasoa, line 57: Instead of first checking with `starts_with` and then striping with `strip_prefix`, we should directly match on the result of strip_prefix:

```suggestion
    if let Some(stripped) = uri.path().strip_prefix(config.prefix().path()) {
        Ok(PathBuf::from(format!("/{}", stripped)))
    } else {
        Err(Redirect(config.prefix().clone()))
    }
```

src/webserver/routing.rs
```diff
@@ -0,0 +1,496 @@
...
+        let target = p.join(NOT_FOUND);
+        if store.contains(&target).await {
+            return NotFound(target);
+        } else {
+            parent = p.parent()
+        }
+    }
+
+    Error(path_to_string(path))
+}
```
lovasoa, line 131:
```suggestion
async fn find_not_found<T>(mut path: &Path, store: &T) -> RoutingAction
where
    T: ExecutionStore,
{
    while let Some(parent) = path.parent() {
        let target = parent.join(NOT_FOUND);
        if store.contains(&target).await {
            return NotFound(target);
        }
        path = parent;
    }

    Error(path_to_string(path))
}

```

src/webserver/routing.rs
```diff
@@ -0,0 +1,496 @@
...
+
+pub trait ExecutionStore {
+    async fn contains(&self, path: &PathBuf) -> bool;
+}
+
+pub trait RoutingConfig {
+    fn prefix(&self) -> &Uri;
+}
+
+pub async fn calculate_route<T, C>(uri: Uri, store: &T, config: &C) -> RoutingAction
```
lovasoa, line 28: why are we taking a Uri ? We should not have to build a Uri to make the routing, we should be able to operate only on the path.

lovasoa:
@guspower , I made a few comments

src/webserver/routing.rs
guspower: good catch, I usually avoid early returns so am trying to use them (and forgetting the consequences!)

src/webserver/routing.rs
guspower: My thinking here was:
1/ trying to increase the guarantee on the input; specifically I don't want to deal with data that isn't a valid uri (encoding issues and what-not).
2/ the contract is uri in to uri out (redirects) or pathbuf out (serve files) - so it felt appropriate
3/ `Uri` makes it easier to use path-only for files and path+query+fragment for redirects (was trying to avoid string splitting and so-forth)

`Uri` itself isn't the most usable or friendly type (it lacks basic stuff like append), but it felt a little better than a plain `&str`.

lovasoa: I think str or bytes are better than building a fake uri just for calling the function. We should not need anything more than a path to make our routing decision.

guspower: ok I can change it; I wasn't particularly wedded to it. 

I had assumed we'd just pass in the value from `actix_web::service::ServiceRequest::uri()` - it seems odd to downcast it to a `&str`?

lovasoa: This uri is reconstructed, and not particularly meaningful (in particular, it is not the uri displayed in the browser's address bar if the service is deployed behind a reverse proxy. I don't think using it is helpful. We won't downcast it, we'll just path the incoming request path unmodified.

guspower: Oh, ok I think I've misunderstood the redirect requirement. I had assumed that we would maintain the query and fragment when redirecting ([as per these tests](https://github.com/guspower/SQLpage/blob/85b4dabf759e1771434118ef520581dc50984e28/src/webserver/routing.rs#L374)). If we don't need this then yep we can just use the `path`.

lovasoa: No, no, you are right ! We should maintain them. We can take a [`PathAndQuery`](https://docs.rs/actix-web/latest/actix_web/http/uri/struct.PathAndQuery.html) as input. This way we do not pass a fake host and scheme, but we don't have to parse a string either.

guspower: Right cool, I'll have a go with `PathAndQuery` this afternoon. Onwards!

guspower:
@lovasoa very very rough first pass at integrating into main_handler. Comment most welcome. The changes to filesystem are too minimal and I need to better understand and integrate `safe_local_path` for existence checking.

src/webserver/http.rs
```diff
@@ -423,41 +425,58 @@ async fn serve_fallback(
...
-        // directeory, and its parent directory, ...
-        Err(e) if e.as_response_error().status_code() == StatusCode::NOT_FOUND => {
-            serve_fallback(&mut service_request, e).await?
+    let app_state: &web::Data<AppState> = service_request.app_data().expect("app_state");
+    let store = AppFileStore::new(&app_state.sql_file_cache, &app_state.file_system);
+    let path_and_query = service_request.uri().path_and_query().expect("expected valid path with query from request");
+    return match calculate_route(path_and_query, &store, &app_state.config).await {
+        NotFound => { serve_fallback(&mut service_request, ErrorWithStatus { status: StatusCode::NOT_FOUND }.into()).await }
+        Execute(path) => { process_sql_request(&mut service_request, path).await }
+        CustomNotFound(path) => { process_sql_request(&mut service_request, path).await }
```
lovasoa, line 434: so we don't actually need a distinct Execute and CustomNotFound ?

guspower: I'm going to work through the flow in `process_sql_request` and come back to you. I'm not sure, but I'm assuming here that the only difference will be in the response status code. It seems logical that the routing will convey found/not found but I will work it through first.

lovasoa:
Thanks a lot @guspower ! I made the required changes and merged. Feel free to open a follow-up PR if you notice I forgot something :)

guspower:
Ah excellent. I'll sync and take a look. w00t! :)

guspower:
Yep cool. Works well. The only outstanding item on my list was whether we should be returning a `404` status code for user-supplied 404 pages?

lovasoa:
I think that would indeed be a nice thing. I didn't add it yesterday because the changes were large already, and that wasn't the behavior of the previous version.

What do you think ? On one hand, people may already have apps that rely on 404.sql for normal pages with non-file-based URLs. On the other hand, people may already be using it to create custom error pages that should have a 404 status code and currently don't.

In any case, we should avoid overriding the status code if the user sets it from their sql code (maybe use RequestContext for that).

guspower:
Yeah I realized there are two different requirements here (I think, maybe more!):

1. Serve up custom "default"/"fallback" pages should the path not have a corresponding file, with status code of `200` (i.e. it's intended behaviour)
2. Serve up actual custom `404` pages with a `404` status code (i.e. it's _really_ not found)

A few possible ways of dealing with it:
- extend `shell` to take a status code (i.e. configure the shell to return `404`)
- support defaulting with both a `404.sql` and a `200.sql` (or `fallback.sql` or some such)
- ignore it :)

It would be really useful to know what people are using the 404.sql for!

lovasoa:
> extend shell to take a status code (i.e. configure the shell to return 404)

We already have a [status_code](https://sql-page.com/component.sql?component=status%5Fcode) component for this.

> support defaulting with both a 404.sql and a 200.sql (or fallback.sql or some such)

I think that's overkill. One can use status_code to switch between the behaviors. The only question is what should be the default one.

---

Another interesting followup would be the ability to configure custom routes, like `"/products/{category}/{id}" : "/product.sql"`

guspower:
Ah yes, custom routes, that was the whole reason I brought this up in the first place lol. Thanks for reminding me :)

I'm traveling for the next couple of days but would like to have a go at a candidate implementation for that later on this week - does that work for you? Will create a new fork.

lovasoa:
Great! And if you could check "allow edits by maintainers" in your PR, that would be great!

Issue #802, opened 2025-02-07
Table sorting does not work with numbers containing thousands separator.


francesco-cattoglio:
**What are you building with SQLPage ?**

A reporting tool with big numbers in a table!

**What is your problem ? A description of the problem, not the solution you are proposing.**

To make the numbers easier to read, I added some formatting with thousands separators, but this breaks column sorting because the sorting internally uses `parseFloat()`, which stops parsing at the first comma it encounters

**Describe the solution you'd like**

The easiest solution would be to strip commas from the string before attempting to parse it as a float. At line 36 of `sqlpage.js` we could replace
```
        return { num: Number.parseFloat(sort_key), str: sort_key };
```
with
```
        return { num: Number.parseFloat(sort_key.replace(/\,/g, "")), str: sort_key };
```
There is a minor drawback, because strings that contain numbers separated by commas (e.g: `15,16,17`) will now be parsed and sorted in a different way. But I do not think that anyone is currently sorting strings like those.

**Describe alternatives you've considered**

I have not considered other alternatives for now. One might argue that my solution is bad since it is locale-dependent, and that is 100% true. But I think it would still be better than current behavior.
A possible flexible solution would be using some kind of markdown, so that the number used for sorting is different from the string used to display it.

francesco-cattoglio:
Hi! I know that in some countries the comma is used as a decimal separator,
since I live in such a country myself! :)
However, the current algorithm for sorting is broken for this usecase
anyway, since right now if the integer part is the same, the rest of the
number is ignored. So, for example, 0,5 and 0,6 are currently rearranged
randomly when sorting.

Il Sab 8 Feb 2025, 21:07 Ophir LOJKINE ***@***.***> ha
scritto:

lovasoa:
fixed by https://github.com/sqlpage/SQLPage/pull/809

thank you !

PR #807, opened 2025-02-11
Added an option to format numbers using `toLocaleString()` for table columns

francesco-cattoglio:
Related to #802

This is the best I could hack, with my limited knowledge and in the time I had to spare. The basic usage is as follows:
- to select which columns should have their numbers formatted according to user locale, you specify them just like you would do for markdown
- there is an extra option that allows you to choose the exact number of decimal digits to be used when formatting, if you need such a precision. This is used for the entire table.

E.G: if you need a table with two columns with numbers formatted using exactly three decimals, and a markdown column with links, you would write
```
select 'table' as component,
    'links'  as markdown,
    'col A' as number_format_locale,
    'col B' as number_format_locale,
    3 as number_format_digits,
    TRUE as sort;
```

**Pros:**
- The changes are very much contained and easy to understand, I do not expect any interference with other features
- Functionality is limited, but can still be useful!

**Cons:**
- *Very* limited functionality. `toLocaleString()` can do so much more, and the list of available options is very very long. I am not sure if we should just allow the user to pass in a JSON structure so that he can choose whatever combination is needed (e.g: this can include formatting with a given currency!)
- no checks w.r.t. the input provided to the `number_format_digits` field, I have not tested to see what happens if a user inputs anything other than an integer
- when this feature is used, the whole table cell contents are replaced. In 99% of cases this should not be a problem, but I could see someone requesting a way to only format the number and keep the rest of the contents (e.g: keep a currency symbol after the number).

Let me know what you think about this!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@
 - Add a new `auto_submit` parameter to the form component. When set to true, the form will be automatically submitted when the user changes any of its fields, and the page will be reloaded with the new value. The validation button is removed.
   - This is useful to quickly create filters at the top of a dashboard or report page, that will be automatically applied when the user changes them.
 - New `options_source` parameter in the form component. This allows to dynamically load options for dropdowns from a different SQL file.
- - This allows easily implementing autocomplete for form fields with a large number of possible options.
- - In the map component, add support for map pins with a description but no title.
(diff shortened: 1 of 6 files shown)


lovasoa:
Thank you very much for this contribution ! I'll take things from here. The results look good, I think I'll revert the logic to be opt out rather than opt in (format numbers by default).

PR #809, opened 2025-02-11
feature/add_markdown_to_forms

theochr72:
add description_md to form component

### DIFF
--- a/sqlpage/templates/form.handlebars
+++ b/sqlpage/templates/form.handlebars
@@ -127,7 +127,7 @@
                         <small class="form-hint mt-0">{{description}}</small>
                     {{/if}}
                     {{#if description_md}}
-                        <small class="form-hint mt-0">{{description_md}}</small>
+                        <small class="form-hint mt-0">{{markdown description_md}}</small>
                     {{/if}}
                 </label>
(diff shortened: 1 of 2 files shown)


sqlpage/templates/form.handlebars
```diff
@@ -31,6 +31,9 @@
                                 {{#if description}}
                                     <small class="form-hint mt-0">{{description}}</small>
                                 {{/if}}
+                                {{#if description_md}}
+                                    <small class="form-hint mt-0">{{description_md}}</small>
```
lovasoa, line 35: We need to render this as markdown, not just text, right? You can use the markdown handlebars helper; see how it's done in other components.

lovasoa (changes requested):
thanks! looks good to me! we just need to render the markdown, not just display it as text

you can also edit one of the examples in the component's documentation to use description_md instead of description

sqlpage/templates/form.handlebars
theochr72: You're right, it should be better now. Let me know what you think! 😊

lovasoa (approved):
Thank you very much for this contribution !

PR #813, opened 2025-02-17
add function sqlpage.client_ip()

Kitsune-Kuro:
All RequestInfos can be called via functions, except client_ip.
I would like to change that.


lovasoa (changes requested):
Thanks! Can we add it to the documentation, and return NULL (None) instead of panicking when the field is absent?

lovasoa:
I made the changes and merged. Thank you !

Issue #782, opened 2025-01-16
show the sqlpage version on the download page


lovasoa:
This page should show what the latest version of sqlpage is: https://sql-page.com/your-first-sql-website/

PR #826, opened 2025-02-22
Insert the newest sqlpage version

Kitsune-Kuro:
For the issue #782.

@lovasoa Feel free to customize it.
At least you have a template.

I used tag_name for the JSON, but name would also work.


lovasoa:
tanks

PR #836, opened 2025-02-28
Option to completely hide the footer from the shell.handlebars template

francesco-cattoglio:
By default handlebars handles all falsy values in the same way. This means that there is no way to easily remove the footer from the shell component, because no matter if you leave it empty or give it a `false` value, the default "Built with SQLPage" one appears.

With this small change, setting the value of the `footer` parameter to the empty string will completely hide the footer. As an example, this is the configuration from the `shell.json` in my project:
```
{
    "component": "shell",
    "footer": "",
    "title": "AA BB CC",
    "javascript": "/scripts/htmx.min.js",
    "link": "/",
    "menu_item": [
        {"link": "p1", "title": "Page 1"},
        {"link": "p2", "title": "Page 2"}
    ]
}

```


lovasoa:
Thanks !

PR #856, opened 2025-03-16
Optional html blocks in markdown

lovasoa:


src/template_helpers.rs
```diff
@@ -251,25 +252,76 @@ fn typeof_helper(v: &JsonValue) -> JsonValue {
...
+impl MarkdownConfig for AppConfig {
+    fn allow_dangerous_html(&self) -> bool {
+        self.markdown_allow_dangerous_html
+    }
+
+    fn allow_dangerous_protocol(&self) -> bool {
+        self.markdown_allow_dangerous_protocol
+    }
+}
+
```
lovasoa, line 269: I think this is overkill

src/template_helpers.rs
```diff
@@ -6,6 +6,7 @@ use handlebars::{
     handlebars_helper, Context, Handlebars, HelperDef, JsonTruthy, PathAndJson, RenderError,
     RenderErrorReason, Renderable, ScopedJson,
 };
+use markdown::Options;
```
lovasoa, line 9: It may be confusing to import something named `Options` globally in [template_helpers.rs](https://github.com/sqlpage/SQLPage/pull/856/files#diff-9ec0c0b9c5d3fca1355e8046e996e5bfc68fd0f61cb76085b26749a5cd2c7b5f) that actually is specific to the markdown helper

src/template_helpers.rs
```diff
@@ -251,25 +252,76 @@ fn typeof_helper(v: &JsonValue) -> JsonValue {
...
+            allow_dangerous_html: false,
+            allow_dangerous_protocol: false,
+        }
+    }
+}
+
 impl MarkdownHelper {
-    fn new(config: &AppConfig) -> Self {
+
+    const ALLOW_UNSAFE: &'static str = "allow_unsafe";
```
lovasoa, line 287: again, may be overkill to define a const if it's just used once...

src/template_helpers.rs
```diff
@@ -251,25 +252,76 @@ fn typeof_helper(v: &JsonValue) -> JsonValue {
...
+            allow_dangerous_html: config.allow_dangerous_html(),
+            allow_dangerous_protocol: config.allow_dangerous_protocol(),
         }
     }
+
+    fn calculate_options(&self, args: &[PathAndJson]) -> Options {
+        let mut options = self.system_options();
+
+        if !options.compile.allow_dangerous_html && args.len() > 1 {
+            let arg = &args[1];
```
lovasoa, line 300: you can use `.get(1)` which is safer

src/template_helpers.rs
```diff
@@ -251,25 +252,76 @@ fn typeof_helper(v: &JsonValue) -> JsonValue {
...
+            if arg.relative_path() == Some(&Self::ALLOW_UNSAFE.to_string()) {
```
lovasoa, line 301: It's strange to use the path instead of the value. This means you cannot pass a variable to your helper. We should probably read the string value of the argument here.

sqlpage/templates/text.handlebars
```diff
@@ -8,7 +8,7 @@
 {{~/if~}}
 {{~#if contents_md~}}
     <div class="remove-bottom-margin {{#if center}}mx-auto{{/if}} {{#if article}}markdown article-text{{/if}}">
-        {{{~markdown contents_md~}}}
+        {{{~markdown contents_md allow_unsafe~}}}
```
lovasoa, line 11: I think the syntax should look like 

```suggestion
        {{{~markdown contents_md 'allow_unsafe'~}}}
```

(pass a string to the helper)

lovasoa (commented):
Thanks for the pr, @guspower ! I made some comments.

lovasoa:
We don't want to take any existing safe code and make it unsafe. If we enable allow_unsafe somewhere, it should be in a new argument, with a big warning in the documentation.

lovasoa:
We should also document the new options to the markdown helper: https://sql-page.com/custom_components.sql

src/template_helpers.rs
guspower: Yes, good catch. I was trying to restrain myself from creating a `markdown` submodule. I'll sort this out now.

src/template_helpers.rs
guspower: This was based on my logging of what was going on with different combinations passed into the handler; I'll switch it to use the variable.

guspower:
> We don't want to take any existing safe code and make it unsafe. If we enable allow_unsafe somewhere, it should be in a new argument, with a big warning in the documentation.

OK can you elaborate a bit further on this point so that I can better understand, specifically around the implementation details? If we use a hard coded string `'allow_unsafe'` rather than reading a variable of the same name, this means that the user will have to create their own templates that supply that value rather than use the existing components? Sorry, I'm probably misunderstanding this point.

guspower:
OK I've updated the implementation and extended the docs. Let me know what you think.

lovasoa:
@guspower , what about this:

```diff
diff --git a/src/template_helpers.rs b/src/template_helpers.rs
index 7be62c9..f6590c6 100644
--- a/src/template_helpers.rs
+++ b/src/template_helpers.rs
@@ -274,8 +274,6 @@ struct MarkdownHelper {
 }
 
 impl MarkdownHelper {
-    const ALLOW_UNSAFE: &'static str = "allow_unsafe";
-
     fn new(config: &impl MarkdownConfig) -> Self {
         Self {
             allow_dangerous_html: config.allow_dangerous_html(),
@@ -283,38 +281,39 @@ impl MarkdownHelper {
         }
     }
 
-    fn calculate_options(&self, args: &[PathAndJson]) -> markdown::Options {
-        let mut options = self.system_options();
-
-        if !options.compile.allow_dangerous_html && args.len() > 1 {
-            if let Some(arg) = args.get(1) {
-                if arg.value().as_str() == Some(Self::ALLOW_UNSAFE) {
-                    options.compile.allow_dangerous_html = true;
-                }
-            }
-        }
-
-        options
-    }
-
-    fn system_options(&self) -> markdown::Options {
+    fn system_options(&self, preset_name: &str) -> Result<markdown::Options, String> {
         let mut options = markdown::Options::gfm();
         options.compile.allow_dangerous_html = self.allow_dangerous_html;
         options.compile.allow_dangerous_protocol = self.allow_dangerous_protocol;
         options.compile.allow_any_img_src = true;
 
-        options
+        match preset_name {
+            "default" => {}
+            "allow_unsafe" => {
+                options.compile.allow_dangerous_html = true;
+                options.compile.allow_dangerous_protocol = true;
+            }
+            _ => return Err(format!("unknown markdown preset: {preset_name}")),
+        }
+
+        Ok(options)
     }
 }
 
 impl CanHelp for MarkdownHelper {
     fn call(&self, args: &[PathAndJson]) -> Result<JsonValue, String> {
-        let options = self.calculate_options(args);
-        let as_str = match args {
-            [v] | [v, _] => v.value(),
-            _ => return Err("expected one or two arguments".to_string()),
+        let (markdown_src_value, preset_name) = match args {
+            [v] => (v.value(), "default"),
+            [v, preset] => (
+                v.value(),
+                preset
+                    .value()
+                    .as_str()
+                    .ok_or("markdown template helper expects a string as preset name")?,
+            ),
+            _ => return Err("markdown template helper expects one or two arguments".to_string()),
         };
-        let as_str = match as_str {
+        let markdown_src = match markdown_src_value {
             JsonValue::String(s) => Cow::Borrowed(s),
             JsonValue::Array(arr) => Cow::Owned(
                 arr.iter()
@@ -326,7 +325,8 @@ impl CanHelp for MarkdownHelper {
             other => Cow::Owned(other.to_string()),
         };
 
-        markdown::to_html_with_options(&as_str, &options)
+        let options = self.system_options(preset_name)?;
+        markdown::to_html_with_options(&markdown_src, &options)
             .map(JsonValue::String)
             .map_err(|e| e.to_string())
     }
```

lovasoa:
This produces more meaningful errors when the helper is not used as expected

guspower:
Yep that looks good to me. My only comment would be to rename `system_options` to something like `get_preset_options` perhaps.

lovasoa:
Thanks @guspower ! It's merged !

PR #858, opened 2025-03-17
Changed docker command to match dockerhub command

remi-coscoy:
Command taken from https://hub.docker.com/r/lovasoa/sqlpage


lovasoa (approved):
Merci Rémi !

Issue #860, opened 2025-03-18
Table footer!


francesco-cattoglio:
**What are you building with SQLPage ?**

Still building the same reporting tool with big numbers in a table 😄 

**What is your problem ? A description of the problem, not the solution you are proposing.**

Since my tables contain a lot of data, it would be nice if it was possible to create a footer at the end of the table with some computed values (e.g. totals or averages)

**Describe the solution you'd like**

I have already hacked a solution by (ab)using the fact that in HTML you can have a `<tfoot>` before a `<tbody>`, and yet everything will be rendered correctly in the end. The solution currently works like this:
- inside the `{{#each_row}}` block in handlebars it is easy to know if this is the first row, but it is not possible to know if this is the last one
- therefore, I have added a top-level parameter called `first_row_is_footer`, so that inside the handlebars template the first row of data is used to build a `<tfoot>` tag. There are a few `ifs` here and there to make sure the rendering is still correct when no footer is present.
- This might seem strange at first, but it actually works out for most situations. I have a temporary table, and before the `select` that shows all table contents, I just do a `select` with some `sum()`s from the same temporary table, so that the code duplication in my SQL files is very little.
- As a bonus, it is easy to also add a `freeze_footers` top level parameter to have the footer sticky as well!
- As a malus, one has to decide if the footer should have the same "number formatting" using the `toLocaleString` function as the table body. For me it is a solid yes, but perhaps in some situations one might not want it, and as far as my changes goes this cannot be choosen user-side

**Additional context**

I have to clean up some code, and perhaps write an example, but if you are interested I might be able to open a pull request by tomorrow!

lovasoa:
Hello !

  I think I'm missing something. Why is there anything special to do to render a footer; why couldn't the footer be a normal table row ?

```sql
select 'table' as component;
select product, quantity, price from bill_items where id=$id;
select 'total' as product, sum(quantity) as quantity, sum(price) as price from my_table where id=$id; -- this is the "footer"; it's just a normal row at the end of the table 
```

aedion-a:
You can test on the example https://sql-page.com/component.sql?component=table (0.34 version)
When the table is sorted, the footer doesn't stay on the last row.
I don't know if it is a bug or a feature.

On the 33.1, I use a workaround to keep sticky footer with a normal row with a specific value in one cell and opacity:0 to hide this value.
With filter, for most use cases, the user doesn't know the "footer" row meets the search criterion and so remains displayed.
[...]

PR #862, opened 2025-03-19
Added an option to create footers in the table component 

francesco-cattoglio:
As per issue #860 this pull request implements two new parameters for the table component: `first_row_is_footer` and `freeze_footers`

### DIFF
--- a/sqlpage/templates/table.handlebars
+++ b/sqlpage/templates/table.handlebars
@@ -93,7 +93,7 @@
                             <td class="text-center">{{default empty_description 'No data'}}</td>
                         </tr>
                     </tbody>
-                {{~/if~}}
+                {{/if}}
             </table>
         </div>
     </div>
--- a/sqlpage/templates/table.handlebars
+++ b/sqlpage/templates/table.handlebars
@@ -58,7 +58,7 @@
                         {{#if ../first_row_is_footer}}
                             {{! If first row is the footer we open the <tfoot> tag, while the <tbody> will be added after this row is rendered }}
                             <tfoot>
-                        {{#else}}
+                        {{else}}
                             {{! If there is no footer we open the <tbody> now, with a 'delay'ed closure }}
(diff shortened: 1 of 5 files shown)


sqlpage/templates/table.handlebars
```diff
@@ -75,15 +85,37 @@
...
                 {{flush_delayed}}
-                {{#if (eq @row_index 0)}}
-                    <tbody class="table-tbody list">
-                        <tr>
-                            <td class="text-center">{{default empty_description 'No data'}}</td>
-                        </tr>
-                    </tbody>
-                {{/if}}
+
+                {{! If not enough rows were rendered, we need to place a 'No data' cell. "Not enough rows" depends on the footer settings }}
```
francesco-cattoglio, line 101: The following `if` block should probably be modified: if can be re-written as a single `if` statement containing the same code. However, the condition is `(first_row_is_footer && row_index<=1) || (!first_row_is_footer && row_index == 0)` which I find unintelligible when written in handlebars, so I am not sure of any preference you might have for this style

lovasoa:
Could we simplify things by introducing a new row-level property `_sqlpage_footer` and asking that the footer comes in last ?

So that we could do 

```
select 'table' as component;

select col1, col2 from my_table;

select true as _sqlpage_footer, sum(col1) as col1, sum(col2) as col2 from my_table;
```

francesco-cattoglio:
Sure thing! I actually think that we could even work around the requirement of the footer being the last row, since any table can have multiple `<tbody>`s and should still work correctly.

francesco-cattoglio:
I implemented the requested change trying to keep the total modifications from the original code as little as possible.
This currently always produces an empty `<tbody></tbody>` when footer is used but everything renders and works correctly AFAICT.
If you want I can hack around a little more to enforce/ensure that the only correct way to use this is to have the `_sqlpage_footer` as the last row, and no empty `<tbody>` sections are created

lovasoa:
thank you @francesco-cattoglio !

PR #897, opened 2025-04-26
Documentation: add "Extensions to SQL" page

ppom0:
Closes #877

I'm quite happy with this addition! I hope it'll save some time to newcomers.

However, I'm not convinced that the page belongs in 'Your first SQL website' directory.
Would you agree to move it up?

Documenting already took quite some time, and I don't have much energy to update this PR if you have suggestions.
So don't hesitate to rephrase and rework things if you don't like them!

PR #903, opened 2025-04-29
Fix SQLite examples having PostgreSQL :: syntax

ppom0:
According to discussion here
https://github.com/sqlpage/SQLPage/pull/897#issuecomment-2835516830


lovasoa (approved):
thank you !

PR #904, opened 2025-04-29
Remove url-encoded surrounding quotes to URL parameters

### LINKED ISSUE #879: sqlpage.link adds percent-encoded double quotes to list parameters.
### Introduction

The `sqlpage.link` function, when passed a list of parameters, serialize the values of the list with their JSON `"`, which are then percent-encoded.

I think they should not be added to the URL.

### To Reproduce

Query this file like this: `/?tags[]=tag_a
```sql
select 'debug' component, $tags tags, sqlpage.link('index.sql', json_object('tags', json_insert($tags, '$[#]', 'new_tag'))) link;
```

### Actual behavior

After following these steps, what happened ?
If you saw an error on the command line or inside your page, then paste it here

```json
{"component":"debug","tags":"[\"tag_a\"]","link":"index.sql?tags[]=%22tag%5Fa%22&tags[]=%22new%5Ftag%22"}
```

### Expected behavior

```json
{"component":"debug","tags":"[\"tag_a\"]","link":"index.sql?tags[]=tag%5Fa&tags[]=new%5Ftag"}
```

### Version information

 - Linux
 - SQLite
 - SQLPage Version: `main`

### Additional context

Here's a dirty patch I wrote that indeed fixes the issue *in my case*:

```diff
diff --git i/src/webserver/database/sqlpage_functions/url_parameter_deserializer.rs w/src/webserver/database/sqlpage_functions/url_parameter_deserializer.rs
index 54caa55..6f87427 100644
--- i/src/webserver/database/sqlpage_functions/url_parameter_deserializer.rs
+++ w/src/webserver/database/sqlpage_functions/url_parameter_deserializer.rs
@@ -59,7 +59,7 @@ impl<'de> Deserialize<'de> for URLParameters {
                             out.encode_and_push(&key);
                             out.0.push_str("[]");
                             out.0.push('=');
-                            out.encode_and_push(&val.to_string());
+                            out.encode_and_push(&val.to_string().replace("\"", ""));
                         }
                     } else {
                         out.push_kv(&key, value);
```

But I suspect a more general solution should be found. Or if only scalars and arrays of scalars are accepted, maybe this is sufficient. I don't know enough about SQLPage internals and assertions to really know.

I'll happily open the PR if you think it's enough, to satisfy my own vanity 😉 

Edit: It can be a bit more sophisticated, removing only quotes at the beginning and end of the file, something along:
```
if val.startswith(") and val.endswith(") then val[1:len-1]
```

ppom0:
Fix #879

Not quite sure of what I'm doing, as I'm discovering the codebase

### DIFF
--- a/src/webserver/database/sqlpage_functions/url_parameter_deserializer.rs
+++ b/src/webserver/database/sqlpage_functions/url_parameter_deserializer.rs
@@ -1,5 +1,6 @@
 use percent_encoding::{percent_encode, NON_ALPHANUMERIC};
 use serde::{Deserialize, Deserializer};
+use serde_json::Value;
 use std::borrow::Cow;
 use std::fmt;
 
@@ -60,13 +61,11 @@ impl<'de> Deserialize<'de> for URLParameters {
                             out.0.push_str("[]");
                             out.0.push('=');
 
-                            let val = val.to_string();
-                            // Remove any surrounding quotes added by serde_json
-                            out.encode_and_push(if val.starts_with("\"") && val.ends_with("\"") {
-                                &val[1..val.len() - 1]
-                            } else {
-                                &val
-                            });
+                            let val = match val {
+                                Value::String(s) => s,
+                                other => other.to_string(),
+                            };
+                            out.encode_and_push(&val);
                         }
                     } else {
                         out.push_kv(&key, value);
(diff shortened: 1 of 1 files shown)


lovasoa:
I think the source of the bug is actually here: https://github.com/sqlpage/SQLPage/blob/main/src%2Fwebserver%2Fdatabase%2Fsqlpage_functions%2Furl_parameter_deserializer.rs#L62

lovasoa:
`val.to_string()` encodes the value as json, so if it's a string, it ends up encoded surrounded with `"`

lovasoa:
sorry for the long delay, and thank you ! the fix will be in 0.35 !

ppom0:
Thanks for actually doing the fix!
I had to dig more to be able to do it and didn't find the time.
Thanks :D

PR #906, opened 2025-05-01
form improvements with new types of field: header and switch

olivierauverlot:
I have added two new types of fields that can be used in a form. With "header", it is possible to group input fields by theme. With "switch", I have added a toggle switch that is more aesthetically pleasing than a traditional radio button.


lovasoa:
Sorry for the delay, and thank you for the contribution, this will be in v0.35 !

Issue #909, opened 2025-05-03
Extend CSP Configuration


guspower:
What are you building with SQLPage ?

A blog

What is your problem ? A description of the problem, not the solution you are proposing.

The Mozilla observatory marks the site down on the CSP configuration. This configuration has not been modified from the default values.

![Image](https://github.com/user-attachments/assets/5a54ab2c-c182-4fb5-b885-77cd4e5183d4)

If I set the configuration value `CONTENT_SECURITY_POLICY` then I lose the nonce functionality as this only kicks in if there is [no explicit configuration specified](https://github.com/sqlpage/SQLPage/blob/main/src/webserver/content_security_policy.rs#L13). 

I also tried to add multiple values to the same header:
```
SELECT 'http_header'          AS component,
       'object-src ''none'';' AS "Content-Security-Policy";

SELECT 'http_header'        AS component,
       'base-uri ''none'';' AS "Content-Security-Policy";
```
but alas it seems the last value wins (they are not combined or multiples sent to the client). 

What are you currently doing ? Since your solution is not implemented in SQLPage currently, what are you doing instead ?

I could workaround the limitation [as described in an earlier issue](https://github.com/sqlpage/SQLPage/issues/467#issuecomment-2589344490). However I would like to extend SQLpage to handle this circumstance.

**Describe the solution you'd like**

The CSP headers could be configured [as per the strict definition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#strict_csp) without workaround:
```
Content-Security-Policy:
  frame-ancestors 'none';
  upgrade-insecure-requests;
  script-src 'nonce-{RANDOM}';
  object-src 'none';
  base-uri 'none';
```

**Describe alternatives you've considered**

It seems like there could be a few ways to handle this:
1. Enhance the CSP configuration parsing to cater for nonce 
2. Go long on CSP configuration and allow each element to be optionally configured ('script-src', 'object-src' etc.)
3. Create a CSP component
4. Make the nonce value available in the SQL context
5. Allow the same header to be written multiple times
6. Combine / append header values with the same name
7. <probably some other ways>

**Additional context**

It would also be great to be able to test a CSP policy change, but looking at it it does seem pretty wild ([report only with reporting endpoints etc.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#testing_your_policy)). There is also the situation where certain pages need a custom CSP configuration because they actually contain an embed.

[...]

PR #911, opened 2025-05-04
Extend CSP Configuration

guspower:
A first pass implementation as discussed in issue #909 

Let me know what you think - I'll update the docs when you're happy with it.


guspower:
Ah I was unaware of the playwright tests - looks like they caught a bug in my implementation. Cool, I will sort it out.

lovasoa (changes requested):
Thank you for taking the time to dive in !

This is a good start, but the logic is a little bit wrong here. Each nonce has to be used only  once. We cannot define a random nonce on startup and then use it repeatedly. Otherwise an attacker can easily make a script injection that contains the known nonce.

How I think it should work:
 - in app_config, we store a parsed csp string. It could be something as simple as `struct CSPFormat { before_nonce: String, after_nonce: String }`
 - add a method to request_context.content_security_policy that takes a `CSPFormat` and generates a header.
 - in render.rs, replace the csp insertion with `request_context.content_security_policy.add_to_response(&mut response, app_state.config.content_security_policy)`

guspower:
Yep makes sense +1. 

I'll take another run at this tomorrow and ensure each nonce is per-request.

guspower:
OK this should now be _correct_ (I added an e2e test to verify). Let me know what you think!

src/webserver/content_security_policy.rs
```diff
@@ -1,40 +1,101 @@
...
 
-impl Default for ContentSecurityPolicy {
-    fn default() -> Self {
-        Self { nonce: random() }
+impl ContentSecurityPolicy {
+    pub fn new<S: Into<String>>(policy: S) -> Self {
+        Self {
+            nonce: random(),
+            policy: policy.into(),
+        }
```
lovasoa, line 22: This copies the CSP template string and reparses it on every request.  I was suggesting to parse it just once at the beginning when we parse the configuration, and then reuse the same parsed string without copying it in memory on every request.

lovasoa (changes requested):

guspower:
Excellent thank you @lovasoa !!!

PR #938, opened 2025-06-13
Added a compile time flag to compile SQLite with builtin math functions

### LINKED ISSUE #937: SQLite built-in math functions are not available in SQLPage
### Introduction

Many built-in mathematical functions are unavailable because the file `sqlite3.c` is compiled without the correct flag.
See https://github.com/rusqlite/rusqlite/issues/1211 as the rationale behind the default compile settings chosen by the upstream library.

### To Reproduce

Any query that uses a function from the page https://www.sqlite.org/draft/lang_mathfunc.html (such as `exp`, `log` and so on) will error out.

### Expected behavior

I think that it would be a sensible default to enable this particular flag at compile time for SQLPage, since some of these functions are very useful in data manipulation.

### Version information

 - Database: the built-in SQLite DB
 - SQLPage Version: 0.35.1

### Additional context

I am opening a pull request with a `config.toml` to compile using the `-DSQLITE_ENABLE_MATH_FUNCTIONS` compile-time option.

francesco-cattoglio:
Closes #937

I think this is the easy way of setting the env var.
If we want to make this optional it could be implemented via features, but IMO making it optional would be overkill.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,7 @@
 [env]
 # To use built-in math functions, this compile time flag must be set
 # See https://www.sqlite.org/draft/lang_mathfunc.html as a reference
-LIBSQLITE3_FLAGS = { value = "-DSQLITE_ENABLE_MATH_FUNCTIONS", force=true }
+# According to Cargo docs this will not overwrite any env var that was already
+# set by the user, and this is a good thing. If the user already set some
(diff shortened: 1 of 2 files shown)


lovasoa:
Thanks! Can you add a test for it?

francesco-cattoglio:
Sure thing!
I will do it as soon as I have some spare time next week. :)

lovasoa:
Merci ! That will be in the next version

PR #941, opened 2025-06-18
Dynamic routing with _.sql

andrewsinnovations:
This update allow you to create a _.sql file that will catch all routing paths in its folder and all subpaths, while still respecting the 404.sql process already in place.

For example, if your folder structure is such:

/
/models
/models/_.sql
/models/example/404.sql

and if you have this code in the /models/_.sql:

```sql
select sqlpage.path();
select sqlpage.variables();
```

and you have this code in /models/example/404.sql:

```sql
select 'Sorry, this page was not found';
```

In your browser, you can access any of these URLs with this result:

1. / - this will render the default "It Works" page because there is no /index.sql
2. /test - This will render the default 404 page built into sqlpage because there is no /404.sql
3. /models/example/test - This will render "Sorry, this page was not found" because of /models/example/404.sql
4. /models/mymodel?id=3 - This will render a page with /models/mymodel as the path, and {id: 3} as variables. It was caught by /models/_.sql
5. /models/mymodel/submodel/test - This will render a page with /models/mymodel/submodel/test as the path, and {} as variables. It was caught by /models/_.sql
6. /models/mymodel/anything/anything_more/yet_even_more/still_going?id=5&val=2 - This will render a page with /models/mymodel/anything/anything_more/yet_even_more/still_going as the path, and {id: 5, val: 2} as variables because it was caught by /models/_.sql

This allows for some simple dynamic routing.

### DIFF
--- a/src/webserver/http.rs
+++ b/src/webserver/http.rs
@@ -533,24 +533,25 @@ pub async fn run_server(config: &AppConfig, state: AppState) -> anyhow::Result<(
         }
         #[cfg(not(target_family = "unix"))]
         anyhow::bail!("Unix sockets are not supported on your operating system. Use listen_on instead of unix_socket.");
-    } else {
-        if let Some(domain) = &config.https_domain {
-            let mut listen_on_https = listen_on;
-            listen_on_https.set_port(443);
-            log::debug!("Will start HTTPS server on {listen_on_https}");
-            let config = make_auto_rustls_config(domain, config);
-            server = server
-                .bind_rustls_0_23(listen_on_https, config)
-                .map_err(|e| bind_error(e, listen_on_https))?;
-        } else if listen_on.port() == 443 {
-            bail!("Please specify a value for https_domain in the configuration file. This is required when using HTTPS (port 443)");
-        }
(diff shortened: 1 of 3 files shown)


andrewsinnovations:
This functionality could possible be achieved with the existing 404.sql functionality. However that method indicates a "file not found" intended use, rather than allowing for dynamic routes. If a file named _.sql is used by convention for a dynamic catch-all, its purpose is more obvious.

lovasoa:
Hi ! 

  Could you walk me through ? The goal is to avoid having to name catch-all files `404.sql`, and allow naming them `_.sql` instead ? If this is only this, then maybe we should instead work on improving the documentation and adding examples, in order to make it easier for people to discover the existing functionality ?

andrewsinnovations:
Yes the intention is solely semantics, with _ and 404 having different meanings. I ran some tests with 404.sql and the functionality is the same. If you'd rather only offer 404 handling for unknown routes, I think that will work and can close the PR.

lovasoa:
Ok ! I have been thinking about ways to improve routing, I'll make my own PR and let you know !

PR #942, opened 2025-06-22
fix docs

astraw:
The `:` character is a colon (not a semi-colon).


lovasoa (approved):
Thanks for the correction!

Issue #949, opened 2025-06-28
login page for SSO


astraw:
**What is your problem ? A description of the problem, not the solution you are proposing.**

As tested with Google OIDC, when a user opens my SQLPage website with SSO for the first time, they are immediately redirected to a "sign in with google" page hosted at https://accounts.google.com/ . This is a little jarring as they expect to see my webpage. Instead, they may think something has gone wrong because they are not at my website but somewhere completely different.

It also may be required in some jurisdictions for the user to actively consent to the use of cookies.

**A clear and concise description of what you want to happen.**

It would be nice if there was an option to create a page for a user for whom SSO auth has not (yet) successfully completed. They can be informed of what will happen when they click something like a "login" button to trigger the redirect. This could also be notified that by proceeding, they are accepting cookies.

alexisrc1:
hello @astraw and thanks for your message, this is not a feature we have on our roadmap yet, but please send me an email at alexis@sql-page.com to discuss this further and see if we can start working on this new functionality !

astraw:
@alexisrc1 this is nothing I need right now but I thought the feedback might be useful.

nip1904:
Hi,
I am very interested in such feature.
Maybe it is possible to add a blacklist option, where you can add an url for bypass the oidc middleware in sqlpage.
So you can define a frontpage url with login link and also come back after logout.

lovasoa:
implemented in [#969](https://github.com/sqlpage/SQLPage/pull/969)

this will be in the next version of sqlpage

PR #969, opened 2025-07-24
Added configuration option to skip OIDC authorization checks for certain endpoints

lenardt-gerhardts:
Added a configuration field called "oidc_skip_endpoints" which is a Vec<String> of endpoints that should be ignored by the OIDC Authentication.

This would also resolve issue https://github.com/sqlpage/SQLPage/issues/949


lovasoa:
Thanks for the pr! that's a very useful feature. 

I think we need to document it better, especially for less technical users who may not be familiar with the term "endpoint", and include examples of it in https://sql-page.com/sso/

Also, I'm wondering whether a prefix whitelist wouldn't be more useful than an individual endpoint blacklist. 

When you have a public site that has a subsection that is reserved to authorized users,  you would configure 

```json
{
    "oidc_protected_path_prefix": ["/private"]
}
```

This way you could put authenticated files in a folder and not have to configure more.

lovasoa:
@lenardt-gerhardts I made a few changes, let me know what you think

lenardt-gerhardts:
Yeah I thought about regex pattern matching for the endpoints, to get rid of problems with assets a colleague of me had.

That being said I really like your idea

We could also implement both and let the user decide, which method they want to use. In my opinion a Whitelist is the better option security wise, but of course both can be misconfigured.

Let me know if you want me to work on the whitelist and if it should contain the regex or not :D

lovasoa:
Yes, good idea. Could you add an `oidc_public_paths` configuration option that has precedence over `oidc_protected_paths`, and document it ?

I think regex configuration is more complicated than needed, and is hard to teach to folks who are not already comfortable with regular expressions.

lenardt-gerhardts:
On it.
 
I would personally like to be able to use wildcards like "/*.private.sql" but i'll leave that decision to you.

lenardt-gerhardts:
I've added the whitelist option. 
Tell me if this implementation is satisfactory for you.

configuration.md
```diff
@@ -25,6 +25,8 @@ Here are the available configuration options and their default values:
...
+| `oidc_protected_paths`                        | `["/"]`                                                      | A list of URL prefixes that should be protected by OIDC authentication. By default, all paths are protected (`["/"]`). If you want to make some pages public, you can restrict authentication to a sub-path, for instance `["/admin", "/users/settings"]`. |
+| `oidc_public_paths`                           | `[]`                                                        | A list of URL prefixes that should be publicly available. By default, no paths are publicly accessible (`[]`). If you want to make some pages public, you can bypass authentication for a sub-path, for instance `["/public/", "/assets/"]`. Keep in mind that without the closing backslashes, that any directory or file starting with `public` or `assets` will be publicly available. This will also overwrite any protected path restriction. If you have a private path `/private` and you define the public path `/private/public/` everything in `/private/public/` will be publicly accessible, while everything else in private will still need authentication. You will not be able to define a private path inside a public path. |
```
lovasoa, line 29:
```suggestion
| `oidc_public_paths`                           | `[]`                                                        | A list of URL prefixes that should be publicly available. By default, no paths are publicly accessible (`[]`). If you want to make some pages public, you can bypass authentication for a sub-path, for instance `["/public/", "/assets/"]`. Keep in mind that without the closing backslashes, that any directory or file starting with `public` or `assets` will be publicly available. This will also overwrite any protected path restriction. If you have a private path `/private` and you define the public path `/private/public/` everything in `/private/public/` will be publicly accessible, while everything else in private will still need authentication.
```

I understand what you meant, but I think it's confusing. When setting `oidc_protected_paths = ["/private"]`, then `/private` is indeed a private path inside a public path (`/` is public).

I think the documentation is clear enough about what is possible and what isn't without this sentence.

Issue #971, opened 2025-07-26
pages handled by a 404.sql should not redirect to add a trailing slash


olivierauverlot:
I am working on a REST API with SQLPage. I noticed a problem with the use of the DELETE verb. If I do not put a `/` character at the end of the URL, I get a 301 Moved Permanently error. I do not have this issue with the other HTTP verbs (GET, PUT, POST, PATCH).

The following code is in the file `/api/v1/users/404.sql`.

```sql
SELECT 'json' AS component;
SELECT 
    CASE sqlpage.request_method()
        WHEN 'GET' THEN 'READ'
        WHEN 'DELETE' THEN 'REMOVE'
    END AS value;
```

You can reproduce the problem with these examples.

```
GET http://localhost:8080/api/v1/users

###
# Error 301 Moved Permanently ?
DELETE http://localhost:8080/api/v1/users/42

###
# works fine !
DELETE http://localhost:8080/api/v1/users/42/
```

lovasoa:
Hello! This is not a bug, and is not specific to any http verb. A GET request has the same behavior.

`/x` is only available when a file named `x.sql` exists.

olivierauverlot:
Hi Lovasoa,

This is not a problem with the resource because it is managed by the code in the file /api/v1/users/404.sql.

The examples trigger the call to the 404.sql file.

The problem is specific to the use of the DELETE verb, which requires a / at the end of the URL, immediately after the resource that must be deleted.

PR #972, opened 2025-07-26
Check for 404 before redirecting on no-extension paths

lovasoa:
The commit improves routing logic by checking if a path would result in
a 404

fixes https://github.com/sqlpage/SQLPage/issues/971 before adding a
trailing slash. This prevents unnecessary redirects when a custom 404
handler exists.


lovasoa:
@Cursor review

PR #1013, opened 2025-09-11
Feature logger component

lenardt-gerhardts:
I have implemented the logger component based on the [Idea #985](https://github.com/sqlpage/SQLPage/discussions/985), with the small change of renaming "output" to "message".

It is not documented yet, since i am still trying to wrap my head around the documentation, but i wanted to know what you think, before committing any further.

### DIFF
--- a/tests/sql_test_files/it_works_log.sql
+++ b/tests/sql_test_files/it_works_log.sql
@@ -1,0 +1,6 @@
+select 'log' as component,
+        'Hello, World!' as message,
+        'info' as priority;
+
+select 'text' as component,
+        'It works !' as contents;


lovasoa:
Good; I'll refactor to avoid the duplication and cloning between RequestInfo and RequestContext, but this is out of scope for this pr.

I think all that is missing before we can merge is 
 - documentation (see for reference `examples/official-site/sqlpage/migrations/06_debug.sql`)
 - a test (maybe just a `tests/sql_test_files/it_works_log.sql`)

lovasoa:
Another thing that would be nice is for the log component to work while in header context. Currently logging something switches to html rendering, which means you can't log anything before the shell.

lenardt-gerhardts:
where can i define or search for a icon the component should get?

lovasoa:
https://tabler.io/icons

PR #1084, opened 2025-11-05
fix ipv6 notation

kryskool:
On Mac OS, when localhost:8080 is defined and sqlpage config file, the SocketAddr is resolved with ipv6 address if ip V6 is available on the internal network.

```json
{
  "listen_on": "localhost:8080"
}
```

On the terminal we can Ctrl + Click, but the URL is incorrect (ipV6 notation use bracket)

Actually we have 

```shell
View your website at:
🔗 http://::1:8080
```

Instead of 

```
View your website at:
🔗 http://[::1]:8080
```

Regards,


lovasoa (approved):
Useful, thanks !

Issue #1070, opened 2025-10-27
Modal top-level parameter 'close' is a reserved keyword in SQL Server


charlieglnp:
### Introduction

The top-level parameter 'close' of the model component is a reserved word in Transect-SQL and results in an error when used on SQL Server.

### To Reproduce

```sql
select  'modal' as component,
		'test' || 'modal' as id,
		'Close' as close;
```
The concatenation is to force the query through SQL server (no error occurs otherwise) 

### Actual behavior

```
In "test.sql": The following error occurred while executing an SQL statement:
error returned from database: Incorrect syntax near the keyword 'close'.
```

### Expected behavior
Modal component to not return an error when 'close' top-level parameter is used

### Version information

 - OS: Windows 11 
 - Database SQL Server 2019
 - SQLPage Version 0.38.0

### Additional context
A workaround is to enclose the top-level parameter in square brackets
```sql
select  'modal' as component,
		'test' || 'modal' as id,
		'Close' as [close];
```

lovasoa:
Hi ! Indeed, SQL [has a huge number of keywords](https://en.wikipedia.org/wiki/List_of_SQL_reserved_words), so we don't forbid reusing them as component properties.

You did find the right workaround: quote parameter names when your db does not allow them unquoted.

We can't remove an existing parameter but we can add aliases. Do you want to make a pull request to sqlpage/templates/modal.handlebars and examples/official-site/sqlpage/migrations/63_modal.sql to add an alias for the "close" parameter ?

charlieglnp:
Would there be any benefit in SQLPage raising a specific error when a property name is a reserved keyword for the connected database? 

I regularly prototype using sqlite and then deploy onto sql server, or sqlite and then onto mariadb and things like this (as well as not knowing some of the more esoteric keywords!) catch me out.

lovasoa:
That would not be trivial to implement correctly, since sqlpage does not know which keywords are reserved on your database... But if you want to give it a try, you're welcome !

charlieglnp:
I walked into that one! I can see there might be short cuts for some databases but others eg SQL Server would rely on a static list. It would be a good excuse to contribute so I will have a play and see where I get to, though it will take some time as I'm starting from scratch.

[...]

PR #1088, opened 2025-11-07
prefix SQL keyword with double undescore

kryskool:
Proof of concept for #1070 to keep the HB templates unmodified,

example

`select 'modal' as component, '123' as id, 'test' as title, 'Close' as __close`

is equivalent to 

`select 'modal' as component, '123' as id, 'test' as title, 'Close' as close`

@lovasoa what do you think ?

### DIFF
--- a/src/webserver/database/sql_to_json.rs
+++ b/src/webserver/database/sql_to_json.rs
@@ -7,13 +7,23 @@ use sqlx::any::{AnyColumn, AnyRow, AnyTypeInfo, AnyTypeInfoKind};
 use sqlx::Decode;
 use sqlx::{Column, Row, TypeInfo, ValueRef};
 
+fn remove_double_underscore(value: &str) -> String {
+    let mut chars = value.chars();
+    chars.next();
+    chars.next();
+    chars.as_str().to_owned()
+}
+
 pub fn row_to_json(row: &AnyRow) -> Value {
     use Value::Object;
 
     let columns = row.columns();
     let mut map = Map::new();
     for col in columns {
-        let key = canonical_col_name(col);
+        let mut key = canonical_col_name(col);
+        if key.starts_with("__") {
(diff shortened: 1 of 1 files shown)


lovasoa:
Thanks. However, I'm not sure adding two underscores is easier than adding quotes around the keyword.

```sql
select 'modal' as component, '123' as id, 'test' as title, 'Close' as "close"
````

Is IMHO just as readable.

Maybe just adding aliases to properties with names that conflict with keywords is the best solution ? So that users don't even have to think about it, can copy-and-paste from the docs and get going whatever database they are using.

**Comment by kryskool** [comment]
It seem double quote on SQL Server is a settings to enable, see [SET QUOTED_IDENTIFIER](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-quoted-identifier-transact-sql?view=sql-server-ver16) 


> Causes SQL Server to follow the ISO rules regarding quotation mark delimiting identifiers and literal strings. 
Identifiers delimited by double quotation marks can be either Transact-SQL reserved keywords or can contain characters not generally allowed by the Transact-SQL syntax rules for identifiers.


`Maybe just adding aliases to properties with names that conflict with keywords is the best solution ?`

i don't know if duplicate properties is a best solution, that's why in this POC i prefer to keep the original property and protect it with double underscore (works for all properties), i can also change to use underscore as prefix and suffix, like

```sql
select 'modal' as component, '123' as id, 'test' as title, 'Close' as _close_
```

lovasoa:
We can already write `'Close' as [close]` without any additional config.

I think aliases are better for the developer. We can just deprecate the old name, even remove it from the documentation in newer versions, and add a new name without conflicts.

kryskool:
no worries about aliases, we can also put a warning on a log when "close" keyword is used on modal and alert components to indicate it's deprecated.

PR #1122, opened 2025-11-22
Adding an alias named contents to the html property of the shell-empty component

olivierauverlot:
If you create a RESTful API with SQLPage and the data format transmitted to the client is not HTML, you can use an alias named `contents` to enhance the readability of your code. This approach is particularly useful for returning data formats such as JSON or XML.

For this use case, using the `html` property is inconsistent and can be confusing when reading the code.


olivierauverlot:
I am glad that we share the same opinion on this matter. I have made the modifications in the documentation :)

lovasoa (approved):

PR #1126, opened 2025-11-24
add delay property to the carousel component

DofSekai:
Specify the delay, in milliseconds, between two images. Useful for set carousel display speed


lovasoa:
Thanks for your contribution !

PR #1134, opened 2025-11-29
Add S3 Functions & Accordion Component

a-gonzalo:
As discussed in  #1120 .

Pending to test and document. First try.

Also, like 1 year ago I started working on an accordion component. I think I didn't make a pull request and I don't know if someone implemented it. I'll check it out and remove the changes if it is done.

### DIFF
--- a/src/webserver/database/sqlpage_functions/s3.rs
+++ b/src/webserver/database/sqlpage_functions/s3.rs
@@ -178,14 +178,9 @@ mod tests {
         // Presigning is a local operation for the most part, but it needs credentials.
         let client = get_s3_client(&config).await;
 
-        let url = get_from_s3_with_client(
-            &config,
-            &client,
-            None,
-            Cow::Borrowed("my-file.txt"),
-        )
-        .await
-        .unwrap();
+        let url = get_from_s3_with_client(&config, &client, None, Cow::Borrowed("my-file.txt"))
+            .await
+            .unwrap();
 
         assert!(url.contains("my-bucket"));
         assert!(url.contains("my-file.txt"));


a-gonzalo:
@lovasoa I think I will need to test it with something like MinIO and make some docs.
Do you miss anything related with this?

Also, I would appreciate if you check the commit for the test "run_all_sql_test_files" [commit1](https://github.com/sqlpage/SQLPage/pull/1134/commits/b6d6fbc41aac5af10c7445d8de279efdc1d793da) [commit2](https://github.com/sqlpage/SQLPage/pull/1134/commits/0b0d9748a00dfa525fd05e736c13147ed55ed491)

lovasoa:
Hi !
  Thanks for taking a stab at this ! 
  This is a large PR and I am currently quite busy so it may take some time, but here are my initial remarks:

 - we need a single well tested and well documented feature per PR
   - in this case we need tests that exercise upload / download against an open source s3 implementation
   - the ci times should not increase too much
 - adding new dependencies is costly, we need some reassurance about: compiled binary size increase, compile time increase, how well the dependencies are maintained
 - the documentation needs to be available on the official site
 - i'm not a huge fan of interpreting the same argument as either a file path or a base64 string. If a file has a name that is a valid base64 string, it can get uploaded instead of base64 data.
 - no custom bespoke security rules. I don't think we should forbid uploading files outside the web root. On the contrary it should be encouraged.
 - I think the user should be able to upload to different buckets let them provide `s3://{bucket}/{key}`, not just key.

kryskool:
Hi @a-gonzalo 

Instead of MinIO (MinIO just move in [maintenance mode](https://github.com/minio/minio/commit/27742d469462e1561c776f88ca7a1f26816d69e2) ) , maybe we can use [RustFS](https://docs.rustfs.com/concepts/comparison.html) 

Regards,

lovasoa:
I'm not sure RustFS is production-ready...

Issue #1183, opened 2026-01-10
Race condition in temp file removal causes panic when multiple instances start simultaneously


mukhtaronif:
### Introduction

When multiple SQLPage instances start simultaneously in the same directory, they encounter a race condition that causes a panic during initialization. This occurs because SQLPage's `create_default_database()` function creates a temporary file to test directory writability, then unconditionally removes it using `.expect()`. When multiple instances try to remove the same file concurrently, approximately 75% of them panic with "No such file or directory" error.

### To Reproduce

1. Create a test directory and navigate to it:
```bash
mkdir sqlpage_test && cd sqlpage_test
```

2. Start multiple SQLPage instances simultaneously in the same directory:
```bash
for i in {1..10}; do
  sqlpage --port $((8080 + i)) &
done
wait
```

3. Observe that approximately 7-8 out of 10 instances crash during startup

### Actual behavior

Approximately 75% of the instances panic during initialization with the following error:

```
thread 'main' panicked at sqlpage-0.41.0/src/app_config.rs:498:52:
removing temp file: Os {
    code: 2,
    kind: NotFound,
    message: "No such file or directory"
}
```

**Race condition sequence:**
1. Instance A creates temp file `sqlpage/sqlpage.db` to test writability
2. Instance B creates the same file (overwrites A's file)
3. Instance A tries to remove the file (line 498)
4. Instance B has already removed it
5. Instance A panics with "No such file or directory"

### Expected behavior

All instances should start successfully without panicking. The temp file creation is only for testing directory writability - if the file no longer exists during removal (because another process removed it), this should be handled gracefully rather than causing a panic.

### Version information

- **OS:** macOS Darwin 24.6.0 (also affects Linux/Unix)
- **Database:** SQLite
- **SQLPage Version:** 0.41.0

### Additional context

**Root cause in source code:**

In `src/app_config.rs`, lines 492-499, the `create_default_database()` function:

```rust
if let Ok(tmp_file) = std::fs::File::create(&default_db_path) {
    log::info!(
        "No DATABASE_URL provided, {} is writable, creating a new database file.",
        default_db_path.display()
    );
    drop(tmp_file);
    std::fs::remove_file(&default_db_path).expect("removing temp file"); // ⚠️ Panics here
    return prefix + &encode_uri(&default_db_path) + "?mode=rwc";
}
```

**Proposed fix:**

Replace the `.expect()` with graceful error handling:

```rust
[...]

PR #1185, opened 2026-01-11
fix: handle temp file removal race condition in concurrent initializa…

mukhtaronif:
…tion

When multiple SQLPage instances start simultaneously in the same directory, they can encounter a race condition during initialization. The create_default_database() function creates a temporary file to test directory writability, then removes it. If multiple instances try to remove the same file concurrently, some panic with 'No such file or directory'.

This commit replaces the .expect() panic with graceful error handling using if let Err(). The writability test has already succeeded by the time we try to remove the file, so whether another instance removed it is irrelevant.

Includes a test that spawns 10 concurrent threads initializing AppConfig to verify no panics occur.

ref #1183

### DIFF
--- a/src/app_config.rs
+++ b/src/app_config.rs
@@ -800,44 +800,4 @@ mod test {
             "Configuration directory should default to ./sqlpage when not specified"
         );
     }
-
-    #[test]
-    fn test_concurrent_initialization_no_panic() {
-        let _lock = ENV_LOCK
-            .lock()
-            .expect("Another test panicked while holding the lock");
-
-        let test_dir = std::env::temp_dir().join(format!(
-            "sqlpage_race_test_{}",
-            std::time::SystemTime::now()
-                .duration_since(std::time::UNIX_EPOCH)
-                .unwrap()
-                .as_nanos()
-        ));
-        std::fs::create_dir_all(&test_dir).unwrap();
-
-        env::remove_var("DATABASE_URL");
-
-        let handles: Vec<_> = (0..10)
-            .map(|_| {
-                let test_dir_clone = test_dir.clone();
-                std::thread::spawn(move || {
-                    let cli = Cli {
-                        web_root: None,
-                        config_dir: Some(test_dir_clone),
-                        config_file: None,
-                        command: None,
-                    };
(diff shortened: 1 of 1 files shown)


src/app_config.rs
```diff
@@ -798,4 +800,44 @@ mod test {
...
+        let test_dir = std::env::temp_dir().join(format!(
+            "sqlpage_race_test_{}",
+            std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap()
+                .as_nanos()
+        ));
+        std::fs::create_dir_all(&test_dir).unwrap();
+
+        env::remove_var("DATABASE_URL");
```
lovasoa: I don't think it's safe to just remove this env var like that. We are actually relying on this env var in other tests when running against a non-default db.

lovasoa:
@mukhtaronif did you actually manage to reproduce the crash with the test ? I tried removing the fix but keeping the test, and it does not catch the problem for me.

mukhtaronif:
> @mukhtaronif did you actually manage to reproduce the crash with the test ? I tried removing the fix but keeping the test, and it does not catch the problem for me.

No, I wasn't able to reliably reproduce the crash with the test either. I also tried reverting the fix and running the test multiple times, and it consistently passed.

The race condition is timing-dependent and difficult to trigger in a test environment. Even with a barrier to synchronize the threads in the test, the file system operations complete so quickly that the race window is extremely narrow.
The test  could still serve as:
  - A regression guard that verifies concurrent initialization doesn't panic (with the fix)
  - Documentation of the expected concurrent usage pattern

mukhtaronif:
We have decided to let surveilr set the DATABASE_URL environment variable before calling AppConfig::from_cli(), which prevents SQLPage from calling create_default_database() entirely and avoids triggering the race condition in the first place.

lovasoa:
ok, let's remove the test altogether then

mukhtaronif:
> ok, let's remove the test altogether then

Sure, Just saw that you already did. We made a change from our end as well and got it to work. Thanks @lovasoa

PR #1227, opened 2026-03-02
Add CLAUDE.md for AI assistant context

theochr72:
I’m adding a claude.md file to help AI tools gain better context about the project.

I recommend using it not only as a development aid, but also to review projects and pull requests with AI. It can potentially add an extra layer of security and reliability.

The claude.md file was fully generated by Claude.

Its purpose is to list the components and functions of the project, helping guide development and enabling more reliable code reviews.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,420 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+SQLPage is a Rust web server that turns `.sql` files into dynamic web applications. Users write SQL queries; SQLPage executes them, maps results to handlebars UI components, and streams HTML to the client.
+
+## Build & Test Commands
+
+```bash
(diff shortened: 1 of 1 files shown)


lovasoa:
Hello, and thank you for contributing !

We use the standard widely supported AGENTS.md instead of the proprietary CLAUDE.md. See
 - https://github.com/sqlpage/SQLPage/blob/main/AGENTS.md
 - https://agents.md/

You can add a claude.md with `@AGENTS.md` for only contents.

Also, the contents you propose seems to be more suited for help building sqlpage applications than for building sqlpage itself.

If you want to point your agent to sqlpage docs for help building apps, you can ask it to read https://sql-page.com/llms.txt

lovasoa (changes requested):
we use the standard agents.md

lovasoa:
we now have a detailed agents.md https://github.com/sqlpage/SQLPage/blob/main/AGENTS.md

PR #1229, opened 2026-03-05
Add margin bottom in the big number component 

wissyfaty:
The big_number component were glued together, but with the class mb-2 , we can now separate the components to create an better user experience .


lovasoa (approved):

lovasoa:
thanks

PR #1274, opened 2026-04-11
Revert "get_path_segment() and is_path_matching() sqlpage functions"

lovasoa:
Reverts sqlpage/SQLPage#1266


lovasoa:
i merged instead of closing, by mistake

PR #1283, opened 2026-05-07
card: add top image attributes for lazy loading

setop:
This will be useful for long image gallery.

Developped according to [HTML specs of img tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#lazy)


lovasoa (approved):
thank you !

examples/official-site/sqlpage/migrations/75_card_top_image_attributes.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO parameter(component, name, description, type, top_level, optional) SELECT 'card', * FROM (VALUES
+    -- item level
+    ('top_image_lasy', 'Whether the top image must be lazily or eagerly loaded. Defaults to false, meaning eagerly', 'BOOLEAN', FALSE, TRUE),
+    ('top_image_width', 'Specify the top image width, in pixel. Help preventing a layout shift', 'INTEGER', FALSE, TRUE),
+    ('top_image_height', 'Specify the top image height, in pixel. Help preventing a layout shift', 'INTEGER', FALSE, TRUE)
+) x;
+
```
lovasoa: can we directly edit the existing migration ? The official site recreates the db from scratch on every deploy anyway

lovasoa (changes requested):

examples/official-site/sqlpage/migrations/75_card_top_image_attributes.sql
setop: Sure

I see your commit in  this way.

Do I have to do anything else?

lovasoa (approved):
we're good to go !

PR #1284, opened 2026-05-10
Fix DuckDB :: casting warning

### LINKED ISSUE #1200: (DuckDB) SQLPage warning when using casting shorthand
### Introduction

SQLPage emits a warning regarding the casting syntax when using `::`.

### To Reproduce

Simplest example without needing any database setup (just a blank DuckDB database):

```sql
select '0'::int
from duckdb_functions()
limit 1;
```

### Actual behavior

Output in the logs:
```
[2026-01-23T22:53:24.247Z WARN  sqlpage::webserver::database::sql] Casting with '::' is not supported on your database. For backwards compatibility with older SQLPage versions, we will transform it to CAST(... AS ...).
```

### Expected behavior

As mentioned in the [DuckDB docs](https://duckdb.org/docs/stable/sql/expressions/cast#explicit-casting):
> _The standard SQL syntax for explicit casting is CAST(expr AS TYPENAME), where TYPENAME is a name (or alias) of one of DuckDB's data types. DuckDB also supports the shorthand expr::TYPENAME, which is also present in PostgreSQL._

As the shorthand `::` is supported, the warning should not appear and there's no need to transform the syntaxt to `CAST(... AS ...)`.

### Version information

 - OS: Debian, but also reproducible on macOS
 - Database: DuckDB
 - SQLPage v0.42.0 (using the `docker.io/lovasoa/sqlpage:main-duckdb` image)

kimjune01:
DuckDB natively supports PostgreSQL's `::` casting shorthand ([docs](https://duckdb.org/docs/stable/sql/expressions/cast#explicit-casting)), but SQLPage was warning users and transforming it to `CAST()`.

This adds `SupportedDatabase::Duckdb` to the allow-list alongside Postgres, Snowflake, and Generic to prevent the unnecessary warning and transformation.

Fixes #1200

### DIFF
--- a/src/webserver/database/sql/parameter_extraction.rs
+++ b/src/webserver/database/sql/parameter_extraction.rs
@@ -572,6 +572,7 @@ impl VisitorMut for ParameterExtractor {
                 ..
             } if ![
                 SupportedDatabase::Postgres,
+                SupportedDatabase::Duckdb,
                 SupportedDatabase::Snowflake,
                 SupportedDatabase::Generic,
             ]


lovasoa (approved):
Thanks you for your contribution!

PR #1312, opened 2026-06-10
Sign the OIDC login flow-state cookie (login CSRF / session fixation)

lovasoa:
# Sign the OIDC login flow-state cookie (login CSRF / session fixation)

The built-in OIDC login flow stored its temporary state in an **unsigned** JSON cookie named after the `state` parameter (`sqlpage_oidc_state_<state>`), and the callback trusted whatever cookie matched the callback's `state` ([oidc.rs:1107 stores unsigned JSON](https://github.com/sqlpage/SQLPage/blob/v0.44.0/src/webserver/oidc.rs#L1107), [oidc.rs:703 selects it by `state`](https://github.com/sqlpage/SQLPage/blob/v0.44.0/src/webserver/oidc.rs#L703), [oidc.rs:716 sets `sqlpage_auth`](https://github.com/sqlpage/SQLPage/blob/v0.44.0/src/webserver/oidc.rs#L716)).

An attacker who can plant a cookie for the SQLPage origin (or a parent domain) primes a `state` + matching `sqlpage_oidc_state_<state>` cookie plus an authorization code for an identity they control. A victim hitting `/sqlpage/oidc_callback?code=<attacker_code>&state=<attacker_state>` then gets `set-cookie: sqlpage_auth=<JWT for the attacker identity>`: login CSRF / session fixation, logging the victim into the attacker's account. (Not token forgery; precondition is cookie-planting.)

## Fix

HMAC-SHA256-sign the cookie value with the OIDC client secret (the same key material as the existing [`compute_logout_signature`](https://github.com/sqlpage/SQLPage/blob/v0.44.0/src/webserver/oidc.rs#L623)) and verify it on the callback. A login-state cookie the server never produced is rejected.

## Proof

New integration test [`test_oidc_forged_state_cookie_is_rejected`](https://github.com/sqlpage/SQLPage/blob/ophir.lojkine/fix-oidc-callback-state-fixation/tests/oidc/mod.rs) drives the callback with an attacker-planted (unsigned) state cookie and asserts no `sqlpage_auth` is set.

Before the fix it fails: the forged cookie produces a login.
```
thread 'oidc::test_oidc_forged_state_cookie_is_rejected' panicked at tests/oidc/mod.rs:466:5:
Forged (un-signed) flow-state cookie must not result in a login. Got cookies: [Cookie { cookie_string: Some("sqlpage_auth=eyJhbG...; ... "), ... }, ...]
```

After the fix, the full OIDC suite passes, including the existing happy-path login:
```
$ cargo test --test mod oidc
running 12 tests
test oidc::test_oidc_forged_state_cookie_is_rejected ... ok
test oidc::test_oidc_happy_path ... ok
...
test result: ok. 12 passed; 0 failed
```

`cargo clippy --all-targets --all-features -- -D warnings` is clean. A focused unit test (`login_flow_state_cookie_round_trips_and_rejects_forgery`) also covers sign/verify, forged-signature rejection, and wrong-secret rejection.

In-flight logins started before the upgrade simply restart the OIDC flow; no config change needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG.md
 
+## Unreleased
+
+- **Security fix: OIDC login CSRF / session fixation.** The temporary login-state cookie used during the built-in OIDC login flow was stored as unsigned JSON. An attacker who could plant a cookie for the SQLPage origin (or a parent domain) was able to pre-seed flow state the server never issued and, together with an authorization code for an identity they control, log a victim's browser into that attacker-controlled identity. SQLPage now signs this cookie with server-held key material (HMAC, keyed by the OIDC client secret) and rejects any login-state cookie it did not produce. You are affected if you use SQLPage's built-in OIDC login (`oidc_issuer_url` is set). No configuration change is needed; just upgrade. Existing logins keep working, and any in-flight login that started before the upgrade simply restarts the OIDC flow.
+
 ## v0.44.0
 
 This release focuses on making production SQLPage apps easier to understand, debug, and operate. Most apps should keep working without SQL changes, but maintainers should review the notes about logging and uploaded-file permissions.
--- a/src/webserver/oidc.rs
+++ b/src/webserver/oidc.rs
@@ -708,7 +708,10 @@ async fn process_oidc_callback(
     let LoginFlowState {
         nonce,
         redirect_target,
-    } = parse_login_flow_state(&tmp_login_flow_state_cookie)?;
+    } = parse_login_flow_state(
+        &tmp_login_flow_state_cookie,
+        &oidc_state.config.client_secret,
+    )?;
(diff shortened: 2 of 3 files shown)


lovasoa:
Closing as out of scope under SQLPage's threat model.

This login-CSRF/session-fixation requires the attacker to plant a cookie on the SQLPage origin, which is the precondition the original report stated. The HMAC-signing approach in this PR does not actually close it: an attacker can start their own legitimate flow, copy the validly server-signed state cookie, and plant it. A server-side one-time state store does not close it either, since the attacker's state is validly issued and the victim consumes it once (that only prevents replay). The effective mitigation would be cookie-injection hardening (e.g. a `__Host-` prefix), but defending against an attacker who can already write cookies to the exact origin is out of scope here.

PR #1354, opened 2026-07-24
Add HTML body alternative to send_mail

### LINKED ISSUE #1353: emails containing HTML
Hi, 

In a future version of SQLPage, it would be really interesting to support sending HTML contents with the `send_mail()` function. 

For example, I need to send account activation emails that include links. At the moment, this isn’t possible. I think this is a very common use case. 

It would be great to have a `body_html` parameter, but also a `body_md` parameter so that the HTML code could be generated from Markdown.

Best regards

lovasoa:
The optional `body_html` field sends a multipart/alternative
message alongside the plain-text `body`, which remains required.

fixes https://github.com/sqlpage/SQLPage/issues/1353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 
  - `sqlpage.send_mail` : add the ability to send HTML-formatted emails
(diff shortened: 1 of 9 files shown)


lovasoa:
@olivierauverlot does that sound like what you meant ?

olivierauverlot:
It’s really cool. It’s perfect for my use case :)

olivierauverlot:
I tested sending HTML emails and it works perfectly.

PR #1363, opened 2026-08-05
New partition component

olivierauverlot:
Provides a mechanism to define custom pagination rules. 

It allows splitting records on the generated page based on a user-specified partitioning criterion,  ensuring controlled data distribution across pages.


olivierauverlot:
partition

olivierauverlot:
image

lovasoa:
The compact control does not reveal the active facet while it is closed: the button says `Classes`, while the active `Constitution` value is only visible after opening the menu.

Mobile screenshot of the compact facet control; Classes is the closed label and Constitution is active only in the opened menu

lovasoa:
Added optional `dropdown_title` for compact facets. Item rows are streamed after the dropdown button, so the template cannot derive the active item title before rendering that button. The query can now supply the selected title explicitly; it defaults to `description` when omitted.

olivierauverlot:
My latest commit is a proposed improvement. The `description` property can now be used in both modes (normal and compact). I kept `dropdown_title`, but it is now used to display text on the selector when no facet has been selected by the user. I also added the `selected_facet` property to indicate the selected facet. This seems more consistent to me and allows for greater flexibility when using the component.

image

image

It’s just a detail, but I’m not a big fan of the name `facet` for this component. I prefer `partition`, which seems more appropriate to me.

lovasoa:
The name **partition** in SQL is generally used to refer to table partitions or window function groups. I think the general term for the UI widget that is implemented in this PR is *facet*.

lovasoa:
Thank you @olivierauverlot :hugs: Your component will be in the next version, to be released today :)

PR #1366, opened 2026-08-08
fix(biome.js) :: fix remaining lint issues

81reap:
### Motivation

- `biome check .` reports 18 warnings on a clean checkout, so a warning introduced by a new change is easy to miss.
- A developer's local `.claude/settings.local.json` is checked too, and fails the run outright.

### Description

- Drop `!important` from the declarations that do not need it.
- Keep it, with a `biome-ignore` naming the reason, in the three places that do: leaflet injects its stylesheet after ours, tabler's spacing utilities are themselves `!important`, and `.markdown pre` caps the height a print rule has to override.
- Use optional chaining in `mdastToDelta` and `open_modal_for_hash`.
- Remove a `biome-ignore` for `noUnusedVariables` that no longer matches anything.
- Drop a `@ts-ignore` on a typed `Buffer.from` call, and give three `Number.parseInt` calls their radix.
- Ignore `.claude/*.json` in `biome.json`, alongside the existing `.zed/*.json`.

### Testing

- `npm test` checks 36 files and reports nothing, where the base reports 18 warnings and 3 infos.
- The browser suite passes 21 tests against a local official site.
- `no console errors on card page` fails here and on the unmodified base alike: the page loads avatars from `avatars.githubusercontent.com`, which this machine cannot verify (`ERR_CERT_AUTHORITY_INVALID`).

---

GitHub cannot base a pull request on a branch that lives in a fork, so all eight target `main` and each one carries the commits of those above it. Review and merge them in order:

1. #1366 :: **fix(biome.js) :: fix remaining lint issues** ← this PR
2. #1367 :: fix(modal) :: give modal component an accessible name
3. #1368 :: fix(map) :: ignore map coordinates that are not a pair of numbers
4. #1369 :: feat(chart) :: render column charts as bar charts
5. #1370 :: fix(chart) :: align stacked series on their X values
6. #1371 :: fix(chart) :: line series up on a category axis for every chart type
7. #1372 :: fix(npm) :: install dependencies only once at root level
8. #1373 :: feat(typescript) :: typecheck browser JavaScript in CI

PR #1367, opened 2026-08-08
fix(modal) :: give modal component an accessible name

81reap:
### Motivation

- The modal sets `aria-labelledby` to its title text rather than to the id of an element, so screen readers announce a dialog with no name.
- The browser test that should have caught this passes `label` to `getByRole`, which Playwright does not accept, so it matched any dialog on the page.

### Description

- Give the modal title an id of `{{id}}-title`, and point `aria-labelledby` at it.
- Use `name` instead of `label` in the two `getByRole` calls of the modal test.
- Add a changelog entry.

### Testing

- The browser suite passes 21 tests, the modal one now looking the dialog up by its accessible name.
- `npm test` checks 36 files and reports nothing.

---

2. #1367 :: **fix(modal) :: give modal component an accessible name** ← this PR


81reap:
re-linking context. @lovasoa this commit is now ready to be reviewed

https://github.com/sqlpage/SQLPage/pull/1366#issuecomment-5241261551

lovasoa (approved):
good catch. thanks !

81reap:
@lovasoa sorry to ping again. this is ready to be closed out. I can raise the next one before heading into work :)

PR #1368, opened 2026-08-08
fix(map) :: ignore map coordinates that are not a pair of numbers

81reap:
### Motivation

- `parseCoords` returns whatever `Number.parseFloat` gives it, so `"48.8"` becomes `[48.8]` and `"north"` becomes `[NaN]`.
- Leaflet rejects such a value, and the exception stops the map being built, so one bad row hides every other marker.

### Description

- `parseCoords` returns `undefined` unless it parsed exactly two finite numbers.
- `createMarker` returns `undefined` for a marker it cannot place, and the caller skips it.
- Add `tests/end-to-end/map-component.spec.ts`, covering a well-formed centre and marker, a centre and a marker with no longitude, and a centre that is not numeric at all.
- Add a changelog entry.

### Testing

- The browser suite passes 26 tests, five of them new.
- The new tests assert that the other markers still appear when one is dropped.
- `npm test` checks 37 files and reports nothing.

---

3. #1368 :: **fix(map) :: ignore map coordinates that are not a pair of numbers** ← this PR


lovasoa (changes requested):
Just ignoring instead of throwing makes the issue harder to debug for the developer, doesn't it? I'd be okay with logging an error to the console instead of blocking all rendering, but I'd rather not silently swallow errors

81reap:
that's a fair call out. I will make the changes and push up a fix once I get home from work

81reap:
turns out i had my laptop on me. pushed up a fix during my lunch break

lovasoa (approved):

PR #1369, opened 2026-08-08
feat(chart) :: render column charts as bar charts

81reap:
### Motivation

- The chart documentation lists `column` as a type, but ApexCharts has no such type, so a `column` chart draws nothing.
- Asking for `stacked` on a type that cannot stack, such as `rangeBar`, leaves the chart empty as well.

### Description

- Resolve the ApexCharts type once, at the top of `build_sqlpage_chart`, through an alias table that maps `column` to `bar`.
- Read that resolved type in the ten places that used to read `data.type`, so an alias behaves like the type it names everywhere.
- Pass `stacked` only for `line`, `area` and `bar`, and ignore it elsewhere.
- Document `column` as a synonym of `bar`, and say which types honour `stacked`.
- Drop a `console.log` of every chart's options.
- Add two changelog entries.

### Testing

- The browser suite passes 28 tests, two of them new: a `column` chart draws vertical bars, and a `rangeBar` chart that asks to be stacked still draws.
- `npm test` checks 38 files and reports nothing.
- `no console errors on card page` fails for the reason described in #1366, on this branch and on the unmodified base alike.

---

GitHub cannot base a pull request on a branch that lives in a fork, so all eight target `main` and each one carries the commits of those above it. Review and merge them in order:

1. #1366 :: fix(biome.js) :: fix remaining lint issues
2. #1367 :: fix(modal) :: give modal component an accessible name
3. #1368 :: fix(map) :: ignore map coordinates that are not a pair of numbers
4. #1369 :: **feat(chart) :: render column charts as bar charts** ← this PR
5. #1370 :: fix(chart) :: align stacked series on their X values
6. #1371 :: fix(chart) :: line series up on a category axis for every chart type
7. #1372 :: fix(npm) :: install dependencies only once at root level
8. #1373 :: feat(typescript) :: typecheck browser JavaScript in CI

Issue #1362, opened 2026-08-05
stacked area graphs / stacked line graphs


81reap:
## What are you building with SQLPage ?

I export my NixOS machine's logs + metrics to parquet using a custom python collector. I use SQLPage + DuckDB in my NixOS config to power dashboards that show me relevant logs + metrics.

## What is your problem ? A description of the problem, not the solution you are proposing.

I would like to have the ability to create stacked area graphs or stacked line graphs. 

## What are you currently doing ? Since your solution is not implemented in SQLPage currently, what are you doing instead ?

Currently for component power draw I am just using a regular line graph, but it makes more sense to be able to see how each component adds up to total power draw. 

---

**Describe the solution you'd like**

I want to be able to create stacked area graphs / stacked line graphs.

**Describe alternatives you've considered**

Not sure there is an alternative. For now I just use a regular line graph.

**Additional context**

example graphs :: https://datavizcatalogue.com/methods/stacked_area_graph.html

81reap:
I had Claude whip up a patch I could apply. this is by no means properly bar raised to be PRed into the codebase but it seems to get close to the general idea. sharing to help maintainers :) 

```sql
select 'chart' as component, 'area' as type, true as stacked,
  'Power' as title, 'watts' as ytitle, true as time, 400 as ymax,
  'blue' as color, 'teal' as color,
  350 as yline, '350 W budget' as yline_label;

with span as (select now() at time zone 'UTC' - interval 6 hour as since)
select ts as x, round(pkg_watts, 1) as y, 'CPU' as series
from cpu_power, span where ts > since
union all
select ts, gpu_power_w, 'GPU' || gpu_index
from gpu_samples, span where ts > since
order by series, x;
```
```patch
--- a/sqlpage/apexcharts.js
+++ b/sqlpage/apexcharts.js
@@ -136,6 +136,31 @@
 
     const chart_type = data.type || "line";
     const options = {
+      // Reference lines are annotations, not series: a series would be stacked
+      // in a stacked chart and filled in an area chart, so it could not mark a
+      // constant. They are drawn only inside the y-axis range, so a line above
+      // the data needs ymax set to keep it on screen.
+      annotations: {
+        yaxis: data.yline.map((y, i) => {
+          // Each entry is merged over ApexCharts' yAxisAnnotation defaults, so
+          // the label key is omitted rather than left undefined: undefined
+          // would replace the default label object instead of falling back to
+          // it.
+          const line = {
[...]

Issue #951, opened 2025-06-30
Error on sorting for chart component


DSMejantel:
### Introduction
Data are not sorting correctly

### To Reproduce

```json
{"component":"debug","title":"Récoltes","type":"bar","stacked":1,"labels":1,"toolbar":1,"xticks":"10","xtitle":"Années"}
{"series":"Miel de Montagne","x":"2020","y":105}
{"series":"Miel  de Forêt","x":"2020","y":40}
{"series":"Miel de Montagne","x":"2021","y":70}
{"series":"Miel de Pissenlit","x":"2022","y":80}
{"series":"Miel de Pissenlit","x":"2023","y":22}
{"series":"Miel de Ronce","x":"2023","y":98}
{"series":"Miel de Ronce","x":"2024","y":193}
```

```sql
SET xticks = (SELECT count(distinct annee) FROM production where rucher_id=$id);
select 
    'chart'    as component,
    'Récoltes' as title,
    'bar'      as type,
    TRUE as stacked,
    TRUE as labels,
    TRUE       as toolbar,
    $xticks as xticks,
    "Années" as xtitle
     where $tab='5';
select 
    categorie as series,
    strftime('%Y',annee) as x,
    sum(total)  as y
    FROM production JOIN miel on production.produit=miel.id where rucher_id=$id and $tab='5' group by strftime('%Y',annee), categorie order by annee;
```

### Screenshots

![Image](https://github.com/user-attachments/assets/a2b08aae-3ff6-4de3-82b6-427d560ad138)

### Version information

 - OS: Ubuntu
 - Database  SQLite
 - SQLPage Version 0.35.2

kryskool:
HI @DSMejantel 

I see this problem sometime (not only in SQLPage), can you try to CAST the "Année" field like:

```diff
SET xticks = (SELECT count(distinct annee) FROM production where rucher_id=$id);
select 
    'chart'    as component,
    'Récoltes' as title,
    'bar'      as type,
    TRUE as stacked,
    TRUE as labels,
    TRUE       as toolbar,
    $xticks as xticks,
    "Années" as xtitle
     where $tab='5';
select 
    categorie as series,
+    CAST(strftime('%Y',annee) as integer) as x,
-    strftime('%Y',annee) as x,
    sum(total)  as y
    FROM production JOIN miel on production.produit=miel.id where rucher_id=$id and $tab='5' group by strftime('%Y',annee), categorie order by annee;
```

Regards,

[...]

Issue #727, opened 2024-12-06
stacked bar graph does not "stack" all data with same label


aedion-a:
### Introduction

The stacked bar graph does not handle well same label.

### To Reproduce

```sql
CREATE TABLE `test` (
  `series` varchar(18) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT NULL,
  `x`      varchar(4)  CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT NULL,
  `value`  bigint(21)  NOT NULL
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;

INSERT INTO `test` (`series`, `x`, `value`) VALUES
('1 an <= x', '1/5', 1),
('x < 6 mois', '1/5', 1),
('x < 6 mois', '3/5', 1),
('x < 6 mois', '4/5', 1),
('1 an <= x', '5/5', 9),
('6 mois <= x < 1 an', '5/5', 9),
('x < 6 mois', '5/5', 14);
COMMIT;
```

```sql
select 
    'chart'   as component,
    'freq' as title,
    'bar'     as type,
    TRUE      as horizontal,
    TRUE      as stacked,
    TRUE      as labels,
    true as toolbar,
    'pink'    as color,
    'azure'   as color,
    'lime'    as color;

select * from test;
```

### Actual behavior

The three "5/5" (three values: 9, 9, 14) are not represented on the same line.

### Screenshots

![pdknhgk5](https://github.com/user-attachments/assets/4f976ce4-e52c-44cd-8ec7-8f97f9e684dc)

### Version information

 - OS: Xubuntu
 - Database: MariaDB 10.3
 - SQLPage Version: 0.31

lovasoa:
Thanks for the report ! The fix will be in 0.32 :)

PR #1370, opened 2026-08-08
fix(chart) :: align stacked series on their X values

81reap:
Addresses https://github.com/sqlpage/SQLPage/issues/1362

### Motivation

- ApexCharts stacks series by point order, not by `x` value, so a series missing a point is squashed on top of the wrong one.
- The chart then incorrect data and totals without any warning.
- SQLPage already aligned series on their `x` values, but only for `bar` charts on a text axis.

### Description

- Order the merged axis by what the series already agree on, falling back to ascending order where they diverge, so one unsorted series cannot reorder the chart.
- Align whenever a chart is stacked, on top of the existing bar-on-a-text-axis case.
- Add `tests/js/`, run by Node's test runner from `npm test`, since this logic is worth testing without a browser.
- Document the behaviour on the `stacked` parameter, and add the stacked area example from #1362.
- Add a changelog entry.

### Testing

- 13 unit tests cover the merge order, dates compared by value, a null treated as missing, and the input left unmutated.
- The browser suite passes 33 tests, five of them new, including a stacked area chart whose series do not share their `x` values.
- `npm test` checks 41 files and reports nothing.

---

5. #1370 :: **fix(chart) :: align stacked series on their X values** ← this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
  - `sqlpage.send_mail` now supports rich email bodies. Use `body_html` for a caller-provided HTML alternative, or `body_md` to render Markdown as HTML. Messages retain a plain-text alternative; `body` may be omitted when `body_md` is used, and `body_md` and `body_html` cannot be combined.
  - Form `options_source` URLs now preserve existing query parameters when adding the dynamic `search` parameter.
  - Map coordinates that are not a pair of numbers, like a latitude with no longitude, are now reported in the browser console and skipped, instead of breaking the whole map.
+ - Stacked charts now stack their series by `x` value instead of by point order, which used to give wrong totals when a series was missing a point.
  - `column` charts now display vertical bars instead of nothing at all.
  - `stacked` is now ignored on chart types that cannot stack, instead of displaying an empty chart.
  - Screen readers now announce the title of the modal component instead of an unnamed dialog.
--- a/examples/official-site/sqlpage/migrations/01_documentation.sql
+++ b/examples/official-site/sqlpage/migrations/01_documentation.sql
@@ -664,7 +664,7 @@ INSERT INTO parameter(component, name, description, type, top_level, optional) S
     ('marker', 'Marker size', 'REAL', TRUE, TRUE),
     ('labels', 'Whether to show the data labels on the chart or not.', 'BOOLEAN', TRUE, TRUE),
     ('color', 'The name of a color in which to display the chart. If there are multiple series in the chart, this parameter can be repeated multiple times.', 'COLOR', TRUE, TRUE),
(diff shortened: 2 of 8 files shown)


build.rs
```diff
@@ -74,6 +74,16 @@ async fn process_input_file(client: &awc::Client, path_out: &Path, original: Fil
...
+        } else if let Some(name) = line
+            .strip_prefix("/* !include ./")
+            .and_then(|rest| rest.strip_suffix(" */"))
+        {
+            let included_path = format!("sqlpage/{name}");
+            println!("cargo:rerun-if-changed={included_path}");
+            let included = std::fs::read(&included_path)
+                .unwrap_or_else(|e| panic!("Unable to read {included_path}: {e}"));
+            outfile.write_all(&included).unwrap();
+            outfile.write_all(b"\n").unwrap();
```
lovasoa: i can review changes to the build system if you want, but not in this PR.

lovasoa (changes requested):

lovasoa:
The `align_categories` function we have is the result of multiple iterations on user feedback. I'm not confident merging this without a good argument that this is not going to break people's existing charts.

What I'd need is a clear description of the difference between the old and the new algorithms and why you think this difference is not going to break any reasonable existing chart.

build.rs
81reap: this change is really only for js files to be able to import other files. I can just refactor it out as an idea for the future instead. I'll keep in the new addition of running frontend tests in the test script

81reap:
I really don't have context on the history of how this function came to be, so I had Claude look into old issues to find out. I've added those examples as tests so that we can check for regression. lmk if you think I may have missed any issues.

| commit | issue | rule it added |
|---|---|---|
| 5ff3b50a | #727 | every series gets a point in every category, and skipped categories count as 0 |
| ca00eaa0 | #930 | a chart with one series keeps the order the query returned |
| ee156d94 | #951 | categories the series don't share are sorted by name |
| 32a875d3 | — | same rules, faster loop |

But to put it in words :: the algorithm is functionally the same; the new part only adds stacked line, area and bar charts whose x is a date or a number.

```js
// before
} else if (categories && chart_type === "bar" && series.length > 1)
// after
} else if (series.length > 1 && (is_stacked || (categories && chart_type === "bar")))
```

Other small changes:

- Points keep their `z` value instead of being rebuilt as `{x, y}`, so nothing visible changes but we do less work. This also means SQLPage should now be able to handle 3D charts but I think that's better scoped for another PR to properly test and fix any other straggling issues in the codebase.
- Two equal dates now match. The old code only ran on strings so it wasn't caught before.

@lovasoa lmk if that makes sense :)

lovasoa:
Thank you !

81reap:
no prob! thanks for the feedback and review :)

PR #1371, opened 2026-08-08
fix(chart) :: line series up on a category axis for every chart type

81reap:
### Motivation

- ApexCharts pairs points by index on a text axis whatever the chart type, not only for `bar`, so a `line` or `scatter` chart whose series skip different labels draws points against the wrong ones.
- Filling the gap with zero, as a stack needs, would be wrong here: it draws a line down to zero where there is simply no measurement.

### Description

- `align_series` takes what a missing point is worth: zero to add nothing to a stack, null to leave a gap.
- Align on a text axis for every chart type but `rangeBar`, whose `y` is a range and whose points ApexCharts already pairs by label.
- Add a changelog entry.

### Testing

- 15 unit tests, the new ones covering the gap left for an unstacked series and a measured zero kept apart from a missing value.
- The browser suite passes 41 tests, eight of them new: `area`, `scatter` and `heatmap` charts lined up on a text axis, a `rangeBar` chart left alone, and a bubble chart keeping its point sizes.
- `npm test` checks 41 files and reports nothing.

---

6. #1371 :: **fix(chart) :: line series up on a category axis for every chart type** ← this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
  - Form `options_source` URLs now preserve existing query parameters when adding the dynamic `search` parameter.
  - Map coordinates that are not a pair of numbers, like a latitude with no longitude, are now reported in the browser console and skipped, instead of breaking the whole map.
  - Stacked charts now stack their series by `x` value instead of by point order, which used to give wrong totals when a series was missing a point.
+ - `line`, `area`, `scatter`, `bubble` and `heatmap` charts with text labels on the x axis now line their series up by label, leaving a gap where a series skips one.
  - `column` charts now display vertical bars instead of nothing at all.
  - `stacked` is now ignored on chart types that cannot stack, instead of displaying an empty chart.
  - Screen readers now announce the title of the modal component instead of an unnamed dialog.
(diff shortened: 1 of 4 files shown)


lovasoa (changes requested):
### code

```sql
SELECT 'chart' AS component, 'treemap' AS type, 'Treemap alignment regression' AS title, 420 AS height;

SELECT 'North America' AS series, 'United States' AS label, 35 AS value;
SELECT 'North America' AS series, 'Canada' AS label, 15 AS value;
SELECT 'Europe' AS series, 'France' AS label, 30 AS value;
SELECT 'Europe' AS series, 'Germany' AS label, 55 AS value;
```

### main

Image

### this pr

Image

the black text on dark grey unreadable text bug is preexisting, but the missing label is new

81reap:
Good catch! I aligned every chart type on a text axis except `rangeBar`, so a `treemap` (has no axis) got its series padded with the other series' labels So North America never ended.

We now use an allowlist for some charts (`line`, `area`, `bar`, `scatter`, `bubble`, `heatmap`) in `align_series_for`. I also added unit tests for your specific example and rebased onto main.

lovasoa (approved):

PR #1373, opened 2026-08-08
feat(typescript) :: typecheck browser JavaScript in CI

81reap:
### Motivation

- The browser scripts carry JSDoc types that nothing checks, so a misspelt property or a missing null check only shows up in a browser.
- The browser tests are TypeScript, but nothing compiles them either. #1367 fixed a `getByRole` option that does not exist; `tsc` reports it as `TS2353`.

### Description

- Add a root `tsconfig.json` running `checkJs` and `strict` over `sqlpage/*.js`, and one per test project.
- Declare in `sqlpage/globals.d.ts` the names the bundle relies on at runtime rather than through an import: `L`, `ApexCharts`, `chart_series`, `TomSelect`, and tabler's bundled bootstrap.
- Declare in `tests/end-to-end/globals.d.ts` what the tests reach for on the page, chiefly the Tom Select instance `sqlpage_select_dropdown` attaches to a `<select>`.
- Fix what the checker found: unguarded `page.goto()` results in the tests, a `querySelectorAll` typed as `Element`, and an `Array.prototype.shift` whose result is only sometimes defined.
- Hoist `@types/node` to the root, add `typescript`, and run `npm run typecheck` from `npm test`.
- Split `table sorting` into three named tests, reading the cells as `td._col_id` and `td._col_Amount_in_stock`. It looked for `td.id` and `td.Amount`, which the table has never rendered, so it sorted two empty lists and passed whatever the table did. The new helper refuses a column of fewer than two cells.

### Testing

- `npm test` checks 46 files, typechecks the three projects and passes the 15 unit tests, all reporting nothing.
- The browser suite passes 43 tests.

---

8. #1373 :: **feat(typescript) :: typecheck browser JavaScript in CI** ← this PR

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,9 @@
         "tests/end-to-end"
       ],
       "devDependencies": {
-        "@biomejs/biome": "^2.5.4"
+        "@biomejs/biome": "^2.5.4",
+        "@types/node": "^26.2.0",
+        "typescript": "^7.0.2"
       }
     },
     "node_modules/@biomejs/biome": {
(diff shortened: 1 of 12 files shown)


81reap:
ah, I thought this change may be a bit controversial since it touched the CI so I kept it last on the stack 😓 maybe that was a bad idea in hindsight

@lovasoa seems like I can rebase and merge from my phone if we follow the stack I of PRs I currently have setup. #1368 would be next.

if you would rather I pop this change off of this branch and rebase it onto mainline as an independent PR you can merge, I will have to wait until I get back home after work to do that

lovasoa:
we are in no hurry, it's going to take some time on my side to process everything too!

81reap:
@lovasoa intentionally left a bug in this CI to show that it would fail. will fix properly once this gets closer to being fully ready for review 

https://github.com/sqlpage/SQLPage/actions/runs/31557036835/job/93991463581?pr=1373

lovasoa (approved):
thx

Issue #1374, opened 2026-08-09
chart: reference lines for limits and for events


81reap:
## What are you building with SQLPage ?

Host telemetry dashboards for a fleet of NixOS machines. Metrics are collected to Parquet, queried with DuckDB, and rendered entirely by SQLPage.

## What is your problem ? A description of the problem, not the solution you are proposing.

While I am able to graph data, it is often lost without additional context of expected limits, ranges, and events. For example
- a CPU temperature against the part's Tjmax
- a power draw against a supply or circuit budget
- a request latency against an SLO
- a disk usage against a capacity alarm

This also holds along the x-axis. A latency chart shows that something
went wrong at 11:30, but not that 10:00 was a deploy and 11:30 to 13:00 was an
incident. Reading it means keeping a second window open.

While it would also be nice for SQLPage to be able to page/alert me or trigger a webhook, I think it is out of scope for this request. See #1350 and #1351 

## What are you currently doing ? Since your solution is not implemented in SQLPage currently, what are you doing instead ?

I am doing a mix of different things. 

1. Emitting y-lines as an extra series to plot. This kinda works, but breaks with `stacked` and `area` charts. 
2. Putting threshold context in the title.

### **Describe the solution you'd like**

I would like to be able to pass in any number of x and y reference lines with the option to upgrade reference lines into reference areas by adding more metadata to the request. Changes should be additive so that it is easy to add/remove reference lines from an end user perspective.

### **Describe alternatives you've considered**

1. Emitting y-lines as an extra series to plot. This kinda works, but breaks with `stacked` and `area` charts. 
2. Putting threshold context in the title.

### **Additional context**

example refrence line :: https://docs.aws.amazon.com/quick/latest/userguide/reference-lines.html

81reap:
I have the changes for this locally, but I will hold off on raising them for now as I currently have quite a few PRs open currently. happy to readjust if maintainers feel differently

https://github.com/sqlpage/SQLPage/issues/1362#issuecomment-5228684573

81reap:
thanks for the help getting this feature merged @lovasoa !!

PR #1375, opened 2026-08-13
feat(chart) :: draw horizontal reference lines

81reap:
### Motivation

#1374 

### Description

- A row with a `yline` is drawn as a line across the chart at that value of the y axis, with `yline_label` and `yline_color` for its text and its color. Reference lines are rows, so a chart can have as many of them as the query returns. A line follows its axis, so on a `horizontal` bar chart a `yline` is drawn down the chart rather than across it. They are not added to the total of a `stacked` chart, and are not filled in an `area` chart.
- Document the three parameters, add two examples (an alarm threshold on a line chart, a quota on a horizontal bar chart), and a changelog entry.

### Testing

- Two new browser tests: one chart draws a line and a label for every `yline` row it receives, and a `horizontal` bar chart draws its `yline` down the chart rather than across it.

---

1. #1371 :: fix(chart) :: line series up on a category axis for every chart type
2. #1375 :: **feat(chart) :: draw horizontal reference lines** ← this PR
3. #1376 :: feat(chart) :: draw vertical reference lines
4. #1377 :: feat(chart) :: turn a reference line into a band

### DIFF
--- a/examples/official-site/sqlpage/migrations/01_documentation.sql
+++ b/examples/official-site/sqlpage/migrations/01_documentation.sql
@@ -687,7 +687,10 @@ INSERT INTO parameter(component, name, description, type, top_level, optional) S
     ('z', 'A third value carried by the point. Used as the bubble radius in a bubble chart, and shown in the tooltip under the name given by the top-level "ztitle".', 'REAL', FALSE, TRUE),
     ('label', 'An alias for parameter "x"', 'REAL', FALSE, TRUE),
     ('value', 'An alias for parameter "y"', 'REAL', FALSE, TRUE),
-    ('series', 'If multiple series are represented and share the same y-axis, this parameter can be used to distinguish between them.', 'TEXT', FALSE, TRUE)
+    ('series', 'If multiple series are represented and share the same y-axis, this parameter can be used to distinguish between them.', 'TEXT', FALSE, TRUE),
+    ('yline', 'Draws a reference line across the chart at this value of the y axis instead of plotting a point, to show a limit such as a quota or an alarm threshold. Not drawn if it falls outside of the axis, so set ymax when the limit is above the data.', 'REAL', FALSE, TRUE),
+    ('yline_label', 'A text to display next to the yline.', 'TEXT', FALSE, TRUE),
+    ('yline_color', 'The name of a color for the yline. Grey by default.', 'COLOR', FALSE, TRUE)
 ) x;
 INSERT INTO example(component, description, properties) VALUES
     ('chart', 'An area chart representing a time series, using the top-level property `time`.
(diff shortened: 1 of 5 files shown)


examples/official-site/sqlpage/migrations/01_documentation.sql
```diff
@@ -675,7 +675,10 @@ INSERT INTO parameter(component, name, description, type, top_level, optional) S
     ('y', 'The value of the point on the vertical axis', 'REAL', FALSE, FALSE),
     ('label', 'An alias for parameter "x"', 'REAL', FALSE, TRUE),
     ('value', 'An alias for parameter "y"', 'REAL', FALSE, TRUE),
-    ('series', 'If multiple series are represented and share the same y-axis, this parameter can be used to distinguish between them.', 'TEXT', FALSE, TRUE)
+    ('series', 'If multiple series are represented and share the same y-axis, this parameter can be used to distinguish between them.', 'TEXT', FALSE, TRUE),
+    ('yline', 'Draws a reference line across the chart at this value of the y axis instead of plotting a point, to show a limit such as a quota or an alarm threshold. Not drawn if it falls outside of the axis, so set ymax when the limit is above the data.', 'REAL', FALSE, TRUE),
+    ('yline_label', 'A text to display next to the yline.', 'TEXT', FALSE, TRUE),
+    ('yline_color', 'The name of a color for the yline. Grey by default.', 'COLOR', FALSE, TRUE)
```
lovasoa: Don't you think it would be better to use common property names ? `color` and `label` that describe the color and the label of the element being drawn, be it a point, a vertical line, a horizontal line, or something else ?

lovasoa (changes requested):
I think we can remove support for yline colors in this pr, then separately add support for both individual point colors and yline colors. What do you think ?

81reap:
@lovasoa I can try to split out the colour part into a separate PR, but I would push back against that and your comment a bit. Currently as the PR stands, users are able to add any number of X or Y reference lines. We would need colour to be able to tell them apart. 

https://github.com/sqlpage/SQLPage/pull/1375/changes#diff-5caa0e48a864e277cf43a80229ed27f3f626e3be5af47fe03f57ff2e27a58105R807-R820

Your comment did bring to mind accessiblity, mainly colour blindness. I can do a follow up PR to add different kinds of dashed reference lines so that end users don't only have to rely on colour to be able to tell the difference.

81reap:
not sure why this merge failed the CI :: https://github.com/sqlpage/SQLPage/pull/1375/changes/b5913d83ddf6ffb56cd3fe8602b21d459f9c65c1

but passed before :: https://github.com/sqlpage/SQLPage/pull/1375/changes/48dbb7d4c8966dbab3d8787f78dd550eb8368c1b

I rebased the branch and pushed it back up to re-try the CI

examples/official-site/sqlpage/migrations/01_documentation.sql
81reap: image

here is the example page that shows how we can currently have multiple y reference lines. lmk you thoughts :) 

https://github.com/sqlpage/SQLPage/pull/1375/changes#diff-5caa0e48a864e277cf43a80229ed27f3f626e3be5af47fe03f57ff2e27a58105R807-R820

lovasoa:
My problem is mainly with having multiple different color row level properties. In my opinion, we should have a single optional row level color property that works for horizontal lines, vertical lines, and individual points. 

`color` is well established and works across components, whereas anything like `yline_color` would be ad hoc and would have to be memorized

81reap:
ohhh you meant like `colour` and have it work for multiple types instead of `yline_colour`. that's a good call. mb I thought you were saying to omit the ability to colour all together

fixed in rev2! collapsed `label` as well

lovasoa (approved):

lovasoa:
Thanks!

PR #1376, opened 2026-08-13
feat(chart) :: draw vertical reference lines

81reap:
### Motivation

#1374 

### Description

- `xline` is `yline` for the x axis
- When `time` is set, an `xline` is a date or a timestamp. On a chart with text labels on the x axis, it is one of those labels.
- `y_reference_lines` generalizes into `reference_lines` to take in rows of x and y reference lines
- Document the three parameters, add a "Marking events" example (a deploy and an incident on a latency chart), and extend the changelog entry.

### Testing

- One new browser test: a time chart draws a line and a label for every `xline` row it receives.

---

1. #1371 :: fix(chart) :: line series up on a category axis for every chart type
2. #1375 :: feat(chart) :: draw horizontal reference lines
3. #1376 :: **feat(chart) :: draw vertical reference lines** ← this PR
4. #1377 :: feat(chart) :: turn a reference line into a band

PR #1377, opened 2026-08-13
feat(chart) :: turn a reference line into a band

81reap:
### Motivation

#1374 

### Description

- Adding `yline_end` or `xline_end` to a reference row turns its line into a band that reaches to that value. Label and color work the same as on a line.
- Update the two documentation examples — the throttling threshold becomes a band up to `ymax`, and the incident on the latency chart covers its whole duration — and extend the changelog entry.

### Testing

- The two reference-line browser tests now each check both shapes: one line and one band on the y axis, and one line and one band on the x axis, with both labels visible.

---

1. #1371 :: fix(chart) :: line series up on a category axis for every chart type
2. #1375 :: feat(chart) :: draw horizontal reference lines
3. #1376 :: feat(chart) :: draw vertical reference lines
4. #1377 :: **feat(chart) :: turn a reference line into a band** ← this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,7 @@
  - `sqlpage.fetch_with_meta` now correctly documents server JSON responses sent under `json_body`, not `body`.
  - Datagrid rows with an icon or image no longer display an unnecessary en-dash placeholder, and an explicitly empty description remains empty.
  - Tooltip title text is now inhertis the same colour as the tooltip text.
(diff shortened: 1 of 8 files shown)


81reap:
@lovasoa seems like we accidentally pulled in a duplicated change log when [mergeing mainline](https://github.com/sqlpage/SQLPage/pull/1401/changes/989e4ae7b06cc1a5160f18a935ed612510896b32) into #1401 . I generally find that always rebasing instead of mergeing avoids these kinds of issues. always happy to rebase my PRs and push back up as well :)

lovasoa (approved):

lovasoa (approved):

PR #1378, opened 2026-08-13
fix(rust) :: format the files `cargo fmt` could not see

81reap:
### Motivation

- `cargo fmt` silently skipped all 38 files under `sqlpage_functions/functions/`. 
- Eleven of those files had drifted from the default style.

### Description

Two commits in this PR. 

- `959f73ad` 
  - moves the module declarations out of the macro into plain `mod` items next to it. A function missing `mod` is a resolution error. An orphaned `mod` is a dead-code warning. 
  - rust-analyzer can now resolve these modules too. 
  - Adds a `.git-blame-ignore-revs` scaffold. We need to wait to merge into mainline to be able to pull the commit hash off mainline to ignore. 
- `a732ccac` runs `cargo fmt --all`. committed in a separate commit so that it can be added to `.git-blame-ignore-revs`.

### Testing

n/a :: no behaviour changes

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,4 @@
+# ignore commits from showing up on git diffs.
+
+# === large formatting commits ===
+# TODO :: add commit once merged into mainline
--- a/src/webserver/database/sqlpage_functions/functions.rs
+++ b/src/webserver/database/sqlpage_functions/functions.rs
@@ -1,14 +1,52 @@
 //! Built-in `SQLPage` SQL functions.
 //!
 //! Every function is a plain `async fn` in its own module under [`functions/`](self). To add one,
-//! create `functions/<name>.rs` with an `async fn <name>` and add it to the
-//! [`sqlpage_functions!`](super::function_traits::sqlpage_functions) call below. The macro declares
-//! the module and adds it to the dispatch enum. Argument conversion and
-//! dispatch are handled generically in [`super::function_traits`].
+//! create `functions/<name>.rs` with an `async fn <name>`, declare the module below and add it to
+//! the [`sqlpage_functions!`](super::function_traits::sqlpage_functions) call. Argument conversion
+//! and dispatch are handled generically in [`super::function_traits`].
 
 use std::fmt::Write;
 
 use super::function_traits::sqlpage_functions;
+
+mod basic_auth_password;
+mod basic_auth_username;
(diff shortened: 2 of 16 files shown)


81reap:
@lovasoa due to the `.git-blame-ignore-revs`, this PR should be merged as is. That way both commits independently show up at the tip of main. If you squash + merge then `fix(functions) :: don't use macro to import and build SqlPageFunctionName` will also be ignored

.git-blame-ignore-revs
```diff
@@ -0,0 +1,4 @@
+# ignore commits from showing up on git diffs.
+
+# === large formatting commits ===
+# TODO :: add commit once merged into mainline
```
lovasoa, line 4: do we need this?

lovasoa (changes requested):
changing the formatting without changing the contributor and agent instructions or the ci does not really make sense to me

src/webserver/database/sqlpage_functions/functions.rs
```diff
@@ -1,15 +1,53 @@
...
+mod set_variable;
+mod uploaded_file_mime_type;
+mod uploaded_file_name;
+mod uploaded_file_path;
+mod url_encode;
+mod user_info;
+mod user_info_token;
+mod variables;
+mod version;
+mod web_root;
```
lovasoa, line 49: no, we need the macro to be able to keep the files in the functions folder and the actual functions in sync. this is documented, we can't change the behavior without changing the instructions

lovasoa (changes requested):

.git-blame-ignore-revs
81reap: yes, this is to ignore the formatting commit so that the blame correctly attributes the person who wrote it rather than the person who formatted it. if you don't care about that I can take it out

src/webserver/database/sqlpage_functions/functions.rs
81reap: there is some prior art to doing it this way :: https://github.com/apache/datafusion/issues/9281

that said I have added a test to check for this assumption so that it pragmatically holds. While I do believe in my engineering skills, I do not consider myself an expert in Rust yet so not sure if there is a better approach here. lmk if you have any ideas, happy to iterate and learn along the way 😄

lovasoa (approved):

Issue #1386, opened 2026-08-16
Improve Codebase :: Contract Drift


81reap:
I have started on this path with #1382 but I thought it would be worth creating an issue to track all of the changes I'm thinking about that will come after this. This audit was done with the help of Claude. 

### B1. `configuration.md` — six defaults are wrong or incomplete

| Option | Documented | Actual | Reference |
|---|---|---|---|
| `listen_on` / `port` | `0.0.0.0:8080` / `8080` | `443` when `https_domain` is set | `app_config.rs:426-436` |
| `database_url` | `sqlite://sqlpage.db?mode=rwc` | path is relative to the configuration directory; falls back to in-memory if that is not writable; a legacy `./sqlpage.db` wins if present | `app_config.rs:630-673` |
| `max_database_pool_connections` | 4 backends listed | 5 branches — ODBC is 50 and in-memory SQLite is 128, neither documented | `connect.rs:82-95` |
| `https_certificate_cache_dir` | `./sqlpage/https` | derived from the process working directory, not from the configured `web_root` or configuration directory | `app_config.rs:711-713` |
| `max_recursion_depth` | "Maximum value is 255" | not a cap — the `u8` type makes 256 a startup failure | `app_config.rs:405-406` |

**Also drifted inside the code.** The Rust doc comment at `app_config.rs:290` says the default upload limit is 10 MiB. The actual default at `:703-705` is 5 MiB, and `configuration.md` correctly says 5 MiB.

**Suggested fix.** Correct the rows. Then, optionally, add a test that parses the (stable, three-column) table and asserts that documented option names match the struct's serde names and that documented defaults match the values produced by deserialising `{}`. 

---

### B2. `sqlpage.request_body()` returns `''` where the documentation says NULL

**What happens.** `examples/official-site/sqlpage/migrations/55_request_body.sql:65` says the function "returns NULL if there is no request body". On a request with no body SQLPage actually returns an empty string.

---

### B3. `fetch_with_meta` emits a `json_body` key that its documentation does not mention

**What happens.** `sqlpage.fetch_with_meta` returns the response body under the key `body` for text responses and under `json_body` for JSON responses. The documentation
(`examples/official-site/sqlpage/migrations/58_fetch_with_meta.sql:14-23`) documents a single `body` key holding "a string, or a json object, depending on the content type".

81reap:
B1 :: #1391 
B2 :: #1392 
B3 :: #1393

lovasoa:
[...]

PR #1391, opened 2026-08-18
fix(config) :: document the real configuration defaults'

81reap:
Addresses B1 in #1386


configuration.md
```diff
@@ -8,13 +8,13 @@ Here are the available configuration options and their default values:
...
-| `listen_on`                                   | 0.0.0.0:8080                                                | Interface and port on which the web server should listen                                                                                                                                                                                               |
-| `database_url` | `sqlite://sqlpage.db?mode=rwc` or `DSN=DuckDB` | Database connection URL, either `dbengine://user:password@host:port/dbname` or an ODBC connection string. Special characters should be [percent-encoded](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding). See [Database connection strings](#database-connection-strings) for details and examples.|
+| `listen_on`                                   | 0.0.0.0:8080, or 0.0.0.0:443 when `https_domain` is set     | Interface and port on which the web server should listen                                                                                                                                                                                               |
```
lovasoa: I do not think this is true. When https_domain is set, but a port number different from 443 is set, SQLPage starts both an HTTP and an HTTPS server. This makes me suspicious about the entire PR. Is this entirely LLM generated, or are these issues you actually encountered while configuring your SQLPage instance ?

lovasoa (changes requested):

configuration.md
81reap: ah, no not LLM generated. I ran into issues when investigating what the config vars do. Claude pointed me to this code, which would change the port if `listen_on` or `port` is `null`.

https://github.com/sqlpage/SQLPage/blob/071d4df03486177e48549632ea1b521dd14e1a6f/src/webserver/http.rs#L658-L676

I forgot to consider that `listen_on` and `port` have default values so they would never be `null`. so your right it binds on both ports.

lovasoa (approved):
thanks

PR #1393, opened 2026-08-18
'fix(fetch_with_meta) :: return json bodies under the documented key

81reap:
addresses B3 in #1386 
> **What happens.** `sqlpage.fetch_with_meta` returns the response body under the key `body` for text responses and under `json_body` for JSON responses. The documentation
(`examples/official-site/sqlpage/migrations/58_fetch_with_meta.sql:14-23`) documents a single `body` key holding "a string, or a json object, depending on the content type".

lmk if you would rather I go the other direction where we document `json_body`. to me it seemed better to overload the known `body` key

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
  - Screen readers now announce the title of the modal component instead of an unnamed dialog.
  - `sqlpage.request_body` and `sqlpage.request_body_base64` now return NULL when the request has no body. A body that cannot be read, such as one exceeding the payload limit, is now reported as an error instead of being silently replaced with an empty body.
  - List-valued configuration options, including OIDC paths and trusted audiences, can now be set through environment variables as space-separated lists.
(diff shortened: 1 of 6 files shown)


lovasoa (changes requested):
I think it's better to avoid breaking existing applications, and avoid polymorphic response fields. Having a json field that can only contain json is easier to handle. But you are right that we should document it!

81reap:
will follow up with a rev2!

81reap:
@lovasoa actually as I was making the change, I realized that documenting it as `json_body` is actually a breaking change when viewed from an end user perspective. https://sql-page.com/functions.sql?function=fetch_with_meta

lmk your thoughts, I can still move forward with `json_body` if you prefer

lovasoa:
I think, by [Hyrum's law](https://www.hyrumslaw.com/), we must keep json_body

81reap:
updated! :)

81reap:
thanks for handling this @lovasoa !

Issue #1389, opened 2026-08-16
Improve Codebase :: Misc


81reap:
I have started on this path with #1382 but I thought it would be worth creating an issue to track all of the changes I'm thinking about that will come after this. This audit was done with the help of Claude. 

Small, independent, and mostly deletions the compiler can verify.

| # | Item | Where |
|---|---|---|
| E1 | Leftover `dbg!(&normalized_path, &self.contents);` in the routing test store. It prints on every lookup in every routing test. It is the only `dbg!` in `src/`. | `routing.rs:640` |
| E2 | `type H0` and `impl CanHelp for H0` — no call sites | `template_helpers.rs:12`, `:523-530` |
| E3 | `impl From<URLParameters> for String` — no call sites | `url_parameters.rs:130-134` |
| E4 | `impl ResponseError for ErrorWithStatus` — unreachable. Every construction site converts to `anyhow::Error` immediately, and the status is recovered by downcast at `error.rs:212`. Its body text never ships. | `error_with_status.rs:20-38` |
| E5 | `impl From<AnyKind> for SupportedDatabase` — reachable only through `otel_name_from_kind`, whose callers all pass a `SupportedDatabase`. It also maps `Odbc → Generic`, which contradicts `DbInfo::database_type`. | `database/mod.rs:115-125` |
| E6 | `strip_site_prefix` is a no-op when called from `serve_file`: `routing.rs:169` already stripped the prefix before `Serve(path)` was built | `http.rs:480-482`, called at `:446` |
| E7 | `telemetry::shutdown_telemetry()` runs only on graceful shutdown. Both `std::process::exit(1)` paths (`main.rs:14`, `:35`) skip it, so traces are lost for exactly the failures traces exist to explain. | `main.rs:53` |
| E8 | Root `test.hurl` is orphaned. Both discovery sites use `find examples -mindepth 2 -maxdepth 2 -name test.hurl`, which cannot match a file at the repository root. | `scripts/test-examples-hurl.sh:74`, `ci.yml:422` |
| E9 | `.dockerignore:9` excludes `mssql/`, which does not exist in the repository. `.dockerignore:3` excludes `sqlpage.db`, but Docker anchors that pattern to the root, so `sqlpage/sqlpage.db` is not excluded and `Dockerfile:43` copies it into the image. | `.dockerignore` |
[...]

PR #1395, opened 2026-08-22
feat(rust) :: add mutants.rs as nonblocking step to CI

81reap:
adds mutation testing for Rust as a non-blocking step to the CI :: https://mutants.rs/

This step is non-blocking as it does find issues, but it is up to the reader to interpret the results (at least for me I don't know of a quick way to do it just yet). It also is kinda really slow. I ran it over the entire codebase on my Ryzen 7 5800X3D ~3h (09:56 → 13:07). I left it running through the day as I was at work and these are the results.

```lock.json
{
  "cargo_mutants_version": "27.1.0",
  "start_time": "2026-08-21T13:56:34.66744901Z",
  "hostname": "jager",
  "username": "reap"
}
```
 see #1396 for a more detailed breakdown of the full run

the goal is to break down these things into groups and create an issue to dicuss more about how this can help improve the codebase. this also is an progammatic way to catch the remaining issues called out in #1389

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -543,3 +543,30 @@ jobs:
           TAG_SUFFIX: ${{ matrix.variant != 'minimal' && format('-{0}', matrix.variant) || '' }}
         run: |
           docker buildx imagetools inspect "${REGISTRY_IMAGE}:${GITHUB_REF_NAME}${TAG_SUFFIX}"
+
+  mutants:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-mutants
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -543,3 +543,30 @@ jobs:
...
+          save-if: ${{ github.ref == 'refs/heads/main' }}
+      - name: Mutate only the lines this pull request changed
+        run: |
+          git diff "origin/${{ github.base_ref }}...HEAD" > mutants.diff
+          cargo mutants --in-diff mutants.diff --jobs 2
+      - uses: actions/upload-artifact@v7
+        if: always()
+        with:
+          name: mutants-report
+          path: mutants.out/
```
81reap, line 572: hmm, I was under the impression that this would also run in the CI so that I could show it as working

lovasoa:
See my comment on the associated issue. Most of what this surfaces sounds like noise to me. 100% test coverage is not a useful goal IMHO. 

New tests are welcome, but each one should come with a reasonable argument for why it is testing an important invariant that could realistically break in the future. Other tests just add churn and slow down development.

PR #1404, opened 2026-08-25
feat(chart) :: add the ability to customize chart data  colours

### LINKED ISSUE #1228: Support for per-bar colors in bar charts
Hello,

  I would like to assign a different color to each bar in a horizontal bar chart, but it seems the color parameter is
  only available at the top-level (component) level and not at the row level.

  Here is what I'm trying to achieve:

```sql
  SELECT
      'chart'                          AS component,
      'Comptes expirant prochainement' AS title,
      'bar'                            AS type,
      TRUE                             AS horizontal,
      TRUE                             AS labels;

  SELECT '30 jours' AS label,
      100           AS value,
      'green'       AS color;

  SELECT '60 jours' AS label,
      200           AS value,
      'orange'      AS color;

  SELECT '90 jours' AS label,
      300           AS value,
      'red'         AS color;
```

  I would expect each bar to have its own color (green, orange, red), but currently only the top-level color parameter
  is applied uniformly to all bars.

  Is there a way to set a different color per row in a single-series bar chart? If not, would it be possible to add
  color as a row-level parameter?

  Thank you!

81reap:
fixes #1228 

Screenshot_2026-08-25_15-33-49


81reap:
missed running lint after adding tests

81reap:
uhhh, not sure why lint is failing. I didn't touch any rust files in this change

lovasoa (approved):

PR #1411, opened 2026-08-31
fix(filesystem) :: return 404 when a request resolves to a directory

81reap:
When a directory name has a file extension it is routed to the static file handler and fails with `IsADirectory` and shows a 500 error page. Now it resolves to `404`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   ```
 
   SQLPage now keeps the variable value, producing `https://api.example.com/john.doe` as expected.
+- A request that resolves to a directory now returns a 404 page. Directory names that contain a dot are routed to the static file handler, which used to fail with a server error instead.
 
 ## v0.46
 
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -90,8 +90,7 @@ impl FileSystem {
                     .await
             }
             (Err(e), _) => {
-                let status = io_error_status(&e)
-                    .unwrap_or(actix_web::http::StatusCode::INTERNAL_SERVER_ERROR);
+                let status = io_error_status(&local_path, &e).await;
                 Err(e).with_status(status).with_context(|| {
                     format!("Unable to read local file metadata for {}", path.display())
                 })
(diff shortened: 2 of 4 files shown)


81reap:
CI failed for unrelated reasons, fixed in #1416

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
  - Chart data rows can set their own `color`, painting a single bar, slice or point instead of the whole series. It applies to `bar`, `column`, `rangeBar`, `pie`, `treemap`, `scatter` and `bubble` charts, and to the markers of a `line` or an `area` chart.
 - Updated the bundled Tabler icon sprite from v3.44.0 to [v3.46.0](https://github.com/tabler/tabler-icons/releases/tag/v3.46.0). This adds 18 icons, including `play-bug`, `remote-control`, `rocking-chair`, `run-sprint`, `tabs`, `treasure-chest`, `vault`, and `yarn`, and includes the upstream fixes listed in the release notes.
 - Recursive `sqlpage.run_sql` errors now show one concise, positioned error instead of repeating the full inclusion chain.
+- A request that resolves to a directory now returns a 404 page. Directory names that contain a dot are routed to the static file handler, which used to fail with a server error instead.
```
lovasoa: same

src/filesystem.rs
```diff
@@ -252,16 +246,23 @@ fn is_path_missing_error(error: &std::io::Error) -> bool {
...
-        ErrorKind::NotFound | ErrorKind::NotADirectory => {
-            Some(actix_web::http::StatusCode::NOT_FOUND)
-        }
-        ErrorKind::PermissionDenied => Some(actix_web::http::StatusCode::FORBIDDEN),
-        _ => None,
+        ErrorKind::NotFound | ErrorKind::NotADirectory => actix_web::http::StatusCode::NOT_FOUND,
+        // Reading a directory reports IsADirectory on unix but PermissionDenied on
+        // windows, so the path itself has to be inspected to tell the two apart.
+        _ if is_local_directory(local_path).await => actix_web::http::StatusCode::NOT_FOUND,
+        ErrorKind::PermissionDenied => actix_web::http::StatusCode::FORBIDDEN,
```
lovasoa, line 259: why `_` ? shouldn't we still have `IsADirectory | PermissionDenied if is_local_directory` to avoid running the check (and the fs access) when this is an unrelated error ?

lovasoa (changes requested):

CHANGELOG.md
81reap: fixed!

src/filesystem.rs
81reap: ah good catch!

lovasoa (approved):

PR #1412, opened 2026-08-31
fix(template_helpers) :: stop plus and minus from overflowing on large integers

81reap:
Both helpers did unchecked `i64` math, so large numbers panicked the renter in debug or silently overflowed in release. Now helpers fall back to floating point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   ```
 
   SQLPage now keeps the variable value, producing `https://api.example.com/john.doe` as expected.
+- The `plus` and `minus` template helpers no longer overflow on large integers.
 - A request that resolves to a directory now returns a 404 page. Directory names that contain a dot are routed to the static file handler, which used to fail with a server error instead.
 - A `content_security_policy` that does not contain `'nonce-{NONCE}'` is now sent as written, instead of being silently dropped and leaving the response with no `Content-Security-Policy` header at all. Setting the option to the empty string still disables the header, as documented.
 
--- a/src/template_helpers.rs
+++ b/src/template_helpers.rs
@@ -103,23 +103,30 @@ fn default_helper(v: &JsonValue, default: &JsonValue) -> JsonValue {
 }
 
 fn plus_helper(a: &JsonValue, b: &JsonValue) -> JsonValue {
-    if let (Some(a), Some(b)) = (a.as_i64(), b.as_i64()) {
-        (a + b).into()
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
  - Chart data rows can set their own `color`, painting a single bar, slice or point instead of the whole series. It applies to `bar`, `column`, `rangeBar`, `pie`, `treemap`, `scatter` and `bubble` charts, and to the markers of a `line` or an `area` chart.
 - Updated the bundled Tabler icon sprite from v3.44.0 to [v3.46.0](https://github.com/tabler/tabler-icons/releases/tag/v3.46.0). This adds 18 icons, including `play-bug`, `remote-control`, `rocking-chair`, `run-sprint`, `tabs`, `treasure-chest`, `vault`, and `yarn`, and includes the upstream fixes listed in the release notes.
 - Recursive `sqlpage.run_sql` errors now show one concise, positioned error instead of repeating the full inclusion chain.
+- The `plus` and `minus` template helpers no longer overflow on large integers.
```
lovasoa: you can't add lines to the changelog of a version that has already been released !

src/template_helpers.rs
```diff
@@ -744,4 +740,153 @@ mod tests {
     fn as_json_context<'a>(path: &'a str, value: &'a Value) -> ScopedJson<'a> {
         ScopedJson::Context(value, vec![path.to_string()])
     }
+
+    mod documented_helpers {
```
lovasoa, line 755: most of these already passed before, right ?

lovasoa (changes requested):
Ideally one pr should address one specific point. If the stated goal is to fix a bug in one particular edge case, then add a test for that particular edge case.

CHANGELOG.md
81reap: fixed!

src/template_helpers.rs
81reap: yep, only prod fails on `integer_arithmetic_falls_back_to_floats_instead_of_overflowing` which also handles

> If the stated goal is to fix a bug in one particular edge case, then add a test for that particular edge case.

81reap:
> Ideally one pr should address one specific point.

the `h0` change seemed too small to PR on it's own and didn't really fit into the stack. I raised it as a separate fix here #1419

lovasoa (approved):

PR #1413, opened 2026-08-31
fix(migrations) :: failed migrations now report the SQL filename

81reap:
Errors from a failed migration now point at the file and the SQL that failed. Before a reversible migration could be reported with the contents of its `.down.sql` half, and any migration with a multi-word name was reported as `0001_add new users.sql` rather than `0001_add_new_users.sql`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   ```
 
   SQLPage now keeps the variable value, producing `https://api.example.com/john.doe` as expected.
+- Errors from a failed migration now point at the file and the SQL that failed. Before a reversible migration could be reported with the contents of its `.down.sql` half, and any migration with a multi-word name was reported as `0001_add new users.sql` rather than `0001_add_new_users.sql`.
 
 ## v0.46
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4603,6 +4603,7 @@ dependencies = [
  "sha2 0.11.0",
  "sqlparser",
  "sqlx-core-oldapi",
+ "tempfile",
  "tokio",
  "tokio-stream",
  "tokio-util",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,6 +140,7 @@ lambda-web = [
 
 [dev-dependencies]
 actix-http = "3"
+tempfile = "3"
 tokio = { version = "1", features = ["rt", "time", "test-util"] }
 
 [build-dependencies]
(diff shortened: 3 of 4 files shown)


81reap:
CI failed for unrelated reasons. fixed in #1416

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
  - Chart data rows can set their own `color`, painting a single bar, slice or point instead of the whole series. It applies to `bar`, `column`, `rangeBar`, `pie`, `treemap`, `scatter` and `bubble` charts, and to the markers of a `line` or an `area` chart.
 - Updated the bundled Tabler icon sprite from v3.44.0 to [v3.46.0](https://github.com/tabler/tabler-icons/releases/tag/v3.46.0). This adds 18 icons, including `play-bug`, `remote-control`, `rocking-chair`, `run-sprint`, `tabs`, `treasure-chest`, `vault`, and `yarn`, and includes the upstream fixes listed in the release notes.
 - Recursive `sqlpage.run_sql` errors now show one concise, positioned error instead of repeating the full inclusion chain.
+- A migration that fails is now always reported with its SQL error, even when it is the only migration in the directory.
```
lovasoa: same

lovasoa (changes requested):
> sqlx only ever reports versions it was given, but MigrateError is non_exhasustive

I do not understand this sentence. Do you have an actual example of a case where an incorrect migration number could be reported ? If so, shouldn't it be fixed in sqlx-oldapi instead ?

81reap:
Maybe I'm reading this wrong, but because we match on version alone this would fail on mainline (`0001_foo.up.sql` + `0001_foo.down.sql`)?

https://github.com/sqlpage/sqlx-oldapi/blob/f8b378d1140348ad90b959d01f88d19dcd2e38f5/sqlx-core/src/migrate/migrator.rs#L153-L170

https://github.com/81reap/SQLPage/blob/f603aa88e52df202c00a64beab9581a9d54b7199/src/webserver/database/migrations.rs#L39-L40

I cleaned up the changes so hopefully they make more sense tho!

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
   ```
 
   SQLPage now keeps the variable value, producing `https://api.example.com/john.doe` as expected.
+- Errors from a failed migration now point at the file and the SQL that failed. Before a reversible migration could be reported with the contents of its `.down.sql` half, and any migration with a multi-word name was reported as `0001_add new users.sql` rather than `0001_add_new_users.sql`.
```
lovasoa, line 16: Oh, i see, thanks!

lovasoa (approved):

PR #1418, opened 2026-09-01
Fix variables in mixed SQLPage expressions

### LINKED ISSUE #1417: Regression in variable management
### Introduction

SQLpage 0.46 : Regression in variable management 

### To Reproduce

Here is a code snippet for MariaDB and PostgreSQL. It uses a .env file to define environment variables.

The .env file :
```
IZLY_TIMEOUT = 10000
IZLY_ENDPOINT = "https://api.example.fr"
IZLY_GRAVITEE_API_KEY = "xxxxxx-xxxxxxxxxxx-xxxxxxxxxx-xxxxxxxxxx"
```

The index.sql file :
```sql
set user_id = 'john.doe';

set izly_request = json_object(
    'timeout_ms', CAST(sqlpage.environment_variable('IZLY_TIMEOUT') AS INTEGER),
    'method', 'GET',
    'url', CONCAT(sqlpage.environment_variable('IZLY_ENDPOINT'), '/', COALESCE($user_id,'')),
    'headers', json_object(
        'Content-Type', 'application/json',
        'X-Gravitee-Api-Key', sqlpage.environment_variable('IZLY_GRAVITEE_API_KEY')
    )
);
SELECT 'debug' AS component, $izly_request;
```

### Actual behavior

It can be observed that the URL is incorrect. The value of the user_id variable is NULL.

```
{"component":"debug","?":"{\"timeout_ms\":10000,\"method\":\"GET\",\"url\":\"https://api.example.fr/\",\"headers\":{\"Content-Type\":\"application/json\",\"X-Gravitee-Api-Key\":\"xxxxxx-xxxxxxxxxxx-xxxxxxxxxx-xxxxxxxxxx\"}}"}
```

### Screenshots

Image

### Expected behavior

The `url` attribute in the JSON data should contain `"https://api.example.fr/john.doe"`.

### Version information

 - OS: Linux et MacOS
 - Database PostgreSQL and MariaDB
 - SQLPage Version 0.46

lovasoa:
## Problem

The expression rewriter used one mutable pass for two different jobs: deciding whether each source expression belonged to the database or SQLPage, and replacing SQLPage values with backend parameter placeholders.

That let a database-looking child be lowered before its parent ownership was known. If a later sibling contained a `sqlpage.*` call, the parent was promoted to per-row SQLPage evaluation. SQLPage expression construction then received the already-lowered child and interpreted its generated `$1`, `?`, or `@p1` placeholder as a source request/SET variable. The lookup returned `NULL` instead of the original variable value.

The unnecessary casts removed in v0.46 had previously hidden this phase-ordering bug by wrapping generated placeholders in opaque database expressions.

## Fix

The rewrite pipeline now has explicit, one-way stages with grep-friendly vocabulary documented in `rewrite/mod.rs`:

- `static_simple_select` plans restricted selects that SQLPage can evaluate without a database row.
- `projection_partitioning` assigns each selected expression to the database or per-row SQLPage evaluation and records private row inputs.
- `database_lowering` creates typed bindings and backend placeholders only after a complete source projection has been partitioned.
- `computed_column_validation` rejects relational clauses that depend on values SQLPage computes after the database returns.
- `sqlpage_expression` owns recognition and construction of SQLPage-evaluated expressions.

Generated placeholders cannot return to projection partitioning. Numbered backends retain source binding order; MySQL and ODBC positional bindings are finalized in rendered placeholder order.

The former 1,177-line rewrite module is split into six focused modules. `sql.rs` plus the rewrite modules now totals 2,018 lines including tests, down from 2,091 before the split.

The execution-plan name `StaticSimpleSelect` now matches the official SQL documentation, which has been corrected to describe the full locally evaluable expression set.

This regression is released as v0.46.1, with the package version and changelog updated accordingly.

## Tests

- Added the exact minimal fixture at `tests/sql_test_files/data/set_variable_lost_in_mixed_expression_nooracle.sql`. Oracle skips this fixture because it renders the incidental numeric database result as `2.0`; the rewrite itself remains covered by database-independent plan tests.
- Added plan-level regression coverage for the lost variable, generated-placeholder phase safety, source and positional binding order, and private row-input JSON flag alignment.
- Ran `cargo fmt --all`.
- Ran `cargo clippy --all-targets --all-features -- -D warnings`.
- Ran `cargo test`: 183 unit tests and 75 integration tests passed.
- Ran the full suite against SQLite, MariaDB, PostgreSQL, and Microsoft SQL Server.
- Oracle and ODBC remain covered by CI because the required local ODBC drivers are unavailable.

Fixes #1417


lovasoa:
## Performance benchmark

Compared base `7eb830b2` with PR head `a3c5467c`, using separate `cargo build --release --bin sqlpage` builds on an AMD Ryzen 7 9700X (8 cores / 16 threads), Linux 6.12.

Workload:

- SQLite in-memory database, one connection, development mode.
- A realistic data-heavy route returning 1,000 rows / 197,253-byte JSON.
- Mixed projections using database expressions together with `sqlpage.url_encode`, `sqlpage.request_method`, `concat`, `coalesce`, JSON construction, and request variables.
- Cold parse: 800 requests per build across 100 distinct identical route files, using the response `Server-Timing: sql_file` measurement.
- Warm serving: 4,000 persistent HTTP/1.1 requests per build after the route was cached.

Results (milliseconds):

| Path | Base median | PR median | Base mean | PR mean | Base p95 | PR p95 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Cold `sql_file` load/rewrite | 0.295 | 0.291 | 0.275 | 0.278 | 0.350 | 0.349 |
| Warm end-to-end request | 47.996 | 47.996 | 47.711 | 47.570 | 52.019 | 52.005 |

The cold rewrite phase is unchanged within measurement noise (median -1.4%, mean +1.1%, p95 -0.3%). Warm end-to-end latency is also unchanged (median 0.0%, mean -0.3%, p95 -0.03%). Both builds returned the same 1,000-row response (same SHA-256: `80ad2fb651237ee4c1cc10c1dcaca020363a1ec4df2b27f2cdad1931a5ea4143`).

**Finding: no measurable performance regression in this realistic SQLite workload.**

PR #1421, opened 2026-09-02
fix(docs) :: remove DSN=DuckDB from the database_url default

81reap:
`create_default_database` only ever builds a `sqlite://` URL, so this isn't a possible default

https://github.com/sqlpage/SQLPage/blob/109b20d14afc0915e1a152e5cae14b633141f0df/src/app_config.rs#L633-L676


lovasoa:
Maybe the docs should be clearer, but dsn=duckdb is the default in our duckdb docker images

81reap:
it seems very unobvious thing to put in the configuration page fro all of SQLPage. furthermore, it seems silly that SQLPage is explaining that it overwrite the default db param for a docker image specifically designed for DuckDB. 

from the perspective of an end user this is an exception to the rule (and not another default) that only applies to consumers of the docker image (and even then, the docker image with `-duckdb` suffix not the regular SQLPage image). I think that the context for overridden defaults for a single docker image should be kept in the description of that docker image. https://hub.docker.com/r/lovasoa/sqlpage

not sure where the Docker description is pulled from tho. it doesn't seem to be codified in the repo

lovasoa:
I am open to a better wording in the documentation, but it must be concise and helpful. We can just state "context-dependent, see below" and add a short section later in the file.

81reap:
I read through the docs again, and I moved it as a line in the `ODBC Connection Strings` section

lovasoa (approved):

PR #1423, opened 2026-09-02
fix(oidc) :: compare logout in constant time

81reap:
Logout URLs are signed with an HMAC over the redirect URI so that an attacker can't forge a logout URL. Verification compared the tags with `!=`, which lowers to a `memcmp` and leaks the leading correct bytes from the response time. `Mac::verify_slice` compares every byte before deciding.

We also end up dropping a base64 encode/decode round trip and a now unreachable "Failed to decode expected signature" branch.


lovasoa (approved):

lovasoa:
thanks!

lovasoa:
` leaks the leading correct bytes from the response time` is still not demonstrated though. Probably better worded as "could theoretically leak ... in ideal conditions"

Issue #1396, opened 2026-08-22
Mutation Testing Results :: Dead Code + Missing Tests


81reap:
results of running mutation testing on the rust codebase. I ran it over the entire codebase on my Ryzen 7 5800X3D ~3h (09:56 → 13:07). I left it running through the day as I was at work and these are the results.

## files
- [caught.txt](https://github.com/user-attachments/files/31325746/caught.txt)
- [lock.json](https://github.com/user-attachments/files/31325745/lock.json)
- [missed.txt](https://github.com/user-attachments/files/31325744/missed.txt)
- [mutants.json](https://github.com/user-attachments/files/31325743/mutants.json)

### issues by file

```bash
[reap@jager:~/Desktop/SQLPage]$ awk -F: 'FNR==NR{m[$1]++;next}{c[$1]++}
  END{for(f in m) printf "%d\t%d\t%d\t%.0f%%\t%s\n", m[f], c[f], m[f]+c[f], 100*m[f]/(m[f]+c[f]), f}' \
  mutants.out/{missed,caught}.txt | sort -k1,1rn -k3,3rn
```

| File | Missed | Caught | Viable | Survival |
|---|---|---|---|---|
| `telemetry.rs` | 52 | 13 | 65 | 80% |
| `template_helpers.rs` | 49 | 26 | 75 | 65% |
| `render.rs` | 29 | 48 | 77 | 38% |
| `webserver/database/sql/rewrite.rs` | 21 | 92 | 113 | 19% |
| `webserver/database/sql_to_json.rs` | 21 | 23 | 44 | 48% |
| `app_config.rs` | 18 | 73 | 91 | 20% |
| `webserver/oidc.rs` | 18 | 54 | 72 | 25% |
| `filesystem.rs` | 18 | 31 | 49 | 37% |
| `webserver/http.rs` | 18 | 26 | 44 | 41% |
| `webserver/lambda_http.rs` | 15 | 3 | 18 | 83% |
| `webserver/database/mod.rs` | 12 | 7 | 19 | 63% |
| `webserver/database/execute_queries.rs` | 10 | 34 | 44 | 23% |
| `webserver/database/csv_import.rs` | 9 | 25 | 34 | 26% |
| `webserver/database/connect.rs` | 9 | 4 | 13 | 69% |
| `webserver/database/sqlpage_functions/function_traits.rs` | 7 | 45 | 52 | 13% |
[...]

PR #1424, opened 2026-09-02
tests(oidc) :: pin auth cookie attributes in test

81reap:
`set_auth_cookie` sets `sqlpage_auth` with `Secure`, `HttpOnly`, `SameSite=Lax` and
`path=/`. No test checks any of them. `test_oidc_happy_path` runs the whole login flow,
but it only checks that the callback redirects and that the next request returns `200`.

The test now reads `sqlpage_auth` from the callback's `Set-Cookie` headers and asserts
all four attributes.

### DIFF
--- a/tests/oidc/mod.rs
+++ b/tests/oidc/mod.rs
@@ -1,6 +1,6 @@
 use actix_web::{
     App, HttpResponse, HttpServer, Responder,
-    cookie::Cookie,
+    cookie::{Cookie, SameSite},
     http::{StatusCode, header},
     test,
     web::{self, Data},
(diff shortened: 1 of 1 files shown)


lovasoa:
empty description = draft

81reap:
this is it, there's nothing more. the PR title is self descriptive. its just improving gaps in testing that was found by #1396. This would have gone with #1423 but following the ask in #1412 I've broken this out into a separate PR since it doesn't really help #1423

tests/oidc/mod.rs
```diff
@@ -430,6 +430,20 @@ async fn test_oidc_happy_path() {
...
+        .find(|c| c.name() == "sqlpage_auth")
+        .expect("a successful login must set the auth cookie");
+    assert_eq!(
+        (
+            auth_cookie.http_only(),
+            auth_cookie.secure(),
+            auth_cookie.same_site(),
+            auth_cookie.path()
+        ),
+        (Some(true), Some(true), Some(SameSite::Lax), Some("/"))
```
lovasoa, line 444: individual asserts would take fewer lines and give better error messages...

lovasoa (changes requested):

tests/oidc/mod.rs
81reap: oh I didn't know that! I thought it was pretty neat that you could assert like this tho, sad its not as par on the debug message. fixed!

lovasoa (approved):

PR #1425, opened 2026-09-02
tests(oidc) :: pin both directions of the discovery staleness test

81reap:
`OidcState::maybe_refresh` does two things. It returns early while the snapshot is
fresh, and it spawns a background refresh once the snapshot is older than `max_age`.
`test_slow_discovery_does_not_block_authenticated_requests` only checked the second
one: advance the clock, make a request, assert the discovery count went up.


lovasoa:
Same, please keep your PRs as draft until ready for review

81reap:
this is it, there's nothing more. the PR title is self descriptive. its just improving gaps in testing that was found by #1396. This would have gone with #1423 but following the ask in #1412 I've broken this out into a separate PR since it doesn't really help #1423

lovasoa (approved):

Issue #1387, opened 2026-08-16
Improve Codebase :: Documentation Drift


81reap:
I have started on this path with #1382 but I thought it would be worth creating an issue to track all of the changes I'm thinking about that will come after this. This audit was done with the help of Claude. 

### C1. `CONTRIBUTING.md` tells contributors to use a parameter type that does not exist

`CONTRIBUTING.md:131` and `:185` instruct contributors to use `'TEXT|BOOLEAN|NUMBER|JSON|ICON|COLOR'` and
`'TEXT|BOOLEAN|NUMBER|JSON'` when documenting a component parameter.

**`NUMBER` is not a valid type.** The `parameter_type` table is seeded with exactly:

```
BOOLEAN, COLOR, HTML, ICON, INTEGER, JSON, REAL, TEXT, TIMESTAMP, URL
```
(`examples/official-site/sqlpage/migrations/01_documentation.sql:11-12`)

`AGENTS.md:134` lists the correct set. `CONTRIBUTING.md:130` also omits `description_md`.

---

### C2. `CONTRIBUTING.md` prescribes checks that are weaker than CI

A contributor who follows `CONTRIBUTING.md` exactly can pass locally and fail CI.

| | `CONTRIBUTING.md` | CI |
|---|---|---|
| Rust lint | `cargo clippy` (`:58-61`) | `cargo clippy --all-targets --all-features -- -D warnings` (`ci.yml:48`) |
| Frontend | `npx @biomejs/biome check .` (`:69`) | `npm test` = `biome check . && npm run typecheck && node --test ...` (`package.json:6-7`) |

Two smaller gaps in the same area:

- `AGENTS.md:91-92` requires a `CHANGELOG.md` entry for user-visible changes. The `CONTRIBUTING.md` pull-request checklist (`:206-211`) never mentions the changelog. 
- `AGENTS.md:131-137` lists six official-site documentation tables but omits `blog_posts` (`12_blog.sql:1-8`).

---

### C3. Component documentation and templates have drifted apart in both directions

The official site stores component documentation as rows in a `parameter` table, spread across roughly 50
`INSERT INTO parameter(...)` statements in 40+ migration files. Nothing links those rows to the `.handlebars` templates they describe. 

**Documented but never implemented.** `text.width` — "How wide the paragraph should be, in characters"
(`01_documentation.sql:149`). There is no `width` in `text.handlebars`, and no handling in `src/`. 

**Implemented but undocumented.** Confirmed by grep:

[...]

Issue #1385, opened 2026-08-16
Improve Codebase :: Defects


81reap:
I have started on this path with #1382 but I thought it would be worth creating an issue to track all of the changes I'm thinking about that will come after this. This audit was done with the help of Claude. 

### A1. Rows are silently dropped after a SQL error on the same page

**What happens.** If a `.sql` file has a statement that fails, and a later statement returns rows without naming a component, those rows disappear without an error.

This page should show two rows. It shows one:

```sql
select 'table' as component;
select 1 as x;
select nonexistent_function();  -- fails
select 2 as x;                  -- expected: rendered. actual: silently dropped
```

**Suggested fix.** Make `Some` mean "a component is open":

- `close_component` uses `take()` instead of `as_mut()`.
- Add an owned `next_component_index` counter to `HtmlRenderContext` rather than reading it off the child.
- Drop the `saved_component` restore. A following row then opens a fresh default component.
---

### A2. `Retry-After` is never sent on connection-pool timeouts

**What happens.** When the database connection pool times out, SQLPage returns `429 Too Many Requests` but never attaches the `Retry-After` header.

**Suggested fix.** Move the `Retry-After` header into the `if let Some(status)` arm, keyed on the status being 429, next to the existing 401 `WWW-Authenticate` handling. Both headers are then functions of the final status.

---

### A3. Asset cache-busting URLs change on every build

**What happens.** The hashed filenames used for `sqlpage.js`, `sqlpage.css` and the other embedded assets are supposed to change only when the asset's contents change. They currently change on **every build**, including rebuilds of the same commit. Rebuilding SQLPage invalidates every browser and CDN cache entry for its assets.

**Suggested fix.** Hash `contents`, which `compress_asset` already has in memory at `:47`, instead of re-reading the compressed file. This also deletes the 20-line streaming read in `hashed_filename`.

---

### A4. Browser initialisation runs twice on every page load

**What happens.** Every function registered with `add_init_fn` runs twice on a normal page load. For most this is idempotent. For forms it is not: every `<input type="file">` with a size limit gets **two** `change` listeners, and every form with `data-auto-submit` gets **two** `change → form.submit()` listeners.

**Suggested fix.** Two independent halves:

1. Make readiness an exclusive choice, not three additive registrations. This alone fixes the doubled listeners.
[...]

PR #1426, opened 2026-09-02
fix(oidc) :: pin the redirect cap on a failed login test

81reap:
When an OIDC callback fails, SQLPage sends the browser back to the identity provider to try again. The `sqlpage_oidc_redirect_count` cookie counts those attempts. Once the count reaches `MAX_OIDC_REDIRECTS`, SQLPage shows the error instead of redirecting again. Without that cap, a misconfigured provider leaves the browser looping between the two.

Adds a test for this case.

### DIFF
--- a/tests/oidc/mod.rs
+++ b/tests/oidc/mod.rs
@@ -263,6 +263,19 @@ fn get_query_param(url: &Url, name: &str) -> String {
         .to_string()
 }
 
+const REDIRECT_COUNT_COOKIE: &str = "sqlpage_oidc_redirect_count";
+const MAX_OIDC_REDIRECTS: u8 = 3;
+
+fn redirect_count_cookie(cookies: &[Cookie<'static>]) -> u8 {
+    cookies
+        .iter()
+        .find(|c| c.name() == REDIRECT_COUNT_COOKIE)
+        .expect("every redirect to the OIDC provider must carry the redirect counter")
+        .value()
+        .parse()
+        .expect("the redirect counter must be a number")
+}
+
 fn permits_storage_after_proxy_adds_freshness(headers: &header::HeaderMap) -> bool {
     // Reproduces https://github.com/sqlpage/SQLPage/issues/1341, where an
     // intermediary added `Cache-Control: max-age=86400` to OIDC 303 responses.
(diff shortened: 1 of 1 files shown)


lovasoa:
switching to draft until the pr is ready (including a motivation)

81reap:
@lovasoa this is it, there's nothing more. the PR title is self descriptive. its just improving gaps in testing that was found by #1396. This would have gone with #1423 but following the ask in #1412 I've broken this out into a separate PR since it doesn't really help #1423

lovasoa:
If you want to change the codebase, you need at least one or two sentences to explain what you want to do and why. This PR name starts with "fix" and does not even actually fix anything ! It's just tests changes. If you think something needs changing in the tests, you need at least one or two sentences to explain what and why.

81reap:
ah i see the confusion. as per conventional commits, `fix` is used to signify a minor version bump which would include updating tests and nothing else. while some specific versions of conventional commits (see Angular) expand this to include test as a prefix I've learned it's bad practice to do that when the base standard doesn't have it

https://github.com/sqlpage/SQLPage/discussions/1402

lovasoa:
I know human written, coherent PR descriptions for each PR introduce some friction, but this is on purpose. A few years ago, just having spent time on writing the code usually demonstrated a high enough level of effort to prove a given change was important enough, and that someone actually cared and had thought deeply about that particular change. But today, this needs to be demonstrated in a separate channel.

81reap:
> But today, this needs to be demonstrated in a separate channel.

I agree. since this is just changes to unit tests I was under the impression that a passing CI would be that separate independent automated check of my code change. 

Alternatively if we want to go back to #1402 to establish new standards to outline the expected separate channels I can def help with that too! I'm just trying to think in the long term of how to avoid these kinds of issues not just with me but others as well :) 

> but this is on purpose

IMO humans and agents, alike, are susceptible to the path of least resistance which leads to [the broken window concept I call out here](https://github.com/sqlpage/SQLPage/issues/1396#issuecomment-5382130337). This is also why I'm choosing to fix the lower hanging fruit of mutation tests before bigger things like #1385, #1386, and #1387. But a written/established standard (even if there are a limited amount of devs right now) helps raise that minimum bar

lovasoa (approved):

PR #1427, opened 2026-09-02
fix(http): Last-Modified is the file's mtime, not now()

### LINKED ISSUE #1323: Static files served with Last-Modified = now() instead of the file mtime
Every static file is served with `Last-Modified` set to `SystemTime::now()` rather than the file's real modification time. The header therefore changes on every request and carries no cache-validation information. There is also no `ETag` or `Cache-Control`, and the local-filesystem and database-filesystem "modified since" comparisons disagree (`>` vs `>=`), so conditional-request behavior differs by backend.

## Reproduction (in-memory SQLite, no other setup)

```sh
mkdir /tmp/sp && cd /tmp/sp
printf 'hello static world\n' > asset.txt
touch -t 202001010000 asset.txt          # file mtime is 2020-01-01
SQLPAGE_DATABASE_URL='sqlite::memory:' sqlpage &
curl -sI http://localhost:8080/asset.txt | grep -i last-modified
sleep 1
curl -sI http://localhost:8080/asset.txt | grep -i last-modified
```

Observed (SQLPage v0.44.1), despite the file mtime being 2020-01-01:

```
last-modified: Fri, 12 Jun 2026 12:02:22 GMT
last-modified: Fri, 12 Jun 2026 12:02:23 GMT
```

The value is "now" and advances by one second between the two requests. A client doing `If-Modified-Since` revalidation can never get a meaningful answer, and caches cannot key on it.

## Root cause

`Last-Modified` is built from `SystemTime::now()` when serving a file: [src/webserver/http.rs#L469](https://github.com/sqlpage/SQLPage/blob/162f996d8bb309aff5a590d95701b4cb0fc3e0c1/src/webserver/http.rs#L469).

The comparison contract also differs between backends: the local filesystem uses strict `>` ([src/filesystem.rs#L254](https://github.com/sqlpage/SQLPage/blob/162f996d8bb309aff5a590d95701b4cb0fc3e0c1/src/filesystem.rs#L254)) while the database filesystem query uses `>=` ([src/filesystem.rs#L303](https://github.com/sqlpage/SQLPage/blob/162f996d8bb309aff5a590d95701b4cb0fc3e0c1/src/filesystem.rs#L303)), so a file modified in the same second as the last check is treated differently depending on where it is stored.

## Suggested fix

Set `Last-Modified` from the actual source mtime (the value is already read for cache validation in `modified_since`), or omit the header when the mtime is unknown. Pick one comparison contract (`>` or `>=`) and use it for both the local and DB filesystems. Optionally add `ETag` / `Cache-Control` for static assets while here.

## Risk and mitigation

Low. Correct `Last-Modified` strictly improves caching and revalidation. The mtime is available for both local files (`fs::metadata`) and DB-backed files (the `last_modified` column on `sqlpage_files`), so no information is missing.

Split out from #1249 (item 3).

vaibhav8a:
## What

`Last-Modified` now carries the file's real modification time, and both filesystems agree on the comparison.

Fixes #1323.

## Verified against your reproduction

```
last-modified: Tue, 31 Dec 2019 18:30:00 GMT     # the file's real mtime, not now()
last-modified: Tue, 31 Dec 2019 18:30:00 GMT     # …and stable 2s later
```

| `If-Modified-Since` | status | |
|---|---|---|
| exactly the mtime | **304** | ✓ |
| older than the mtime | **200** | ✓ |
| newer than the mtime | **304** | ✓ |

## How

`FileSystem::modified_at` reads the mtime through the **same local-then-database fallback** `read_file` uses, so the timestamp always describes the file actually served. `DbFsQueries` gains a prepared statement for the `last_modified` column, mirroring the existing `was_modified` / `read_file` / `exists` trio.

When no mtime is available the header is **omitted** rather than filled with a guess — your issue offers that option and I think it is the right one: a header saying "now" is worse than an absent header, because a client believes it.

## On the comparison contract: `>`, and I got this wrong first

You said pick one and use it for both. I picked `>=` initially, reasoning it was the "safe" side.

**Driving your reproduction proved that wrong.** `If-Modified-Since: T` asks whether the file changed *after* T, so an mtime of exactly T is not a change and must answer 304. Under `>=` every revalidation re-sent the whole body — defeating the header this change exists to repair. The `304` row in the table above was a `200` until I fixed it.

So both sides now use `>`, which is the *correct* half rather than merely the chosen one. The DB query moves from `>=` to `>`; local stays `>`.

## Scope

I did **not** add `ETag` or `Cache-Control`. Your issue lists them as optional, and they are a separate decision about caching policy rather than a correctness fix — happy to follow up if you want them.

## Checks

- `cargo build` / `cargo check` — clean
- `cargo clippy --all-targets` — 0 warnings
- `cargo fmt --check` — clean
- `cargo test --lib` — 191 passed

(Local tests needed `unixodbc` installed to link; that is a machine setup detail, not a code change.)

### DIFF
--- a/src/filesystem.rs
+++ b/src/filesystem.rs
@@ -98,6 +98,50 @@ impl FileSystem {
         }
     }
 
+    /// When the file was last modified, or `None` when nothing can say.
+    ///
+    /// Same local-then-database fallback as `modified_since`, because a file served from
+    /// one must report the mtime of that same one. Used for the `Last-Modified` header,
+    /// which was previously `SystemTime::now()` - a value that changed on every request
+    /// and so carried no cache-validation information at all.
+    ///
+    /// `None` rather than a guess when the mtime is unavailable: the caller omits the
+    /// header instead, which is honest. A header that says "now" is worse than no header,
+    /// because a client believes it.
+    pub(crate) async fn modified_at(
+        &self,
+        app_state: &AppState,
+        access: FileAccess<'_>,
+    ) -> Option<DateTime<Utc>> {
+        let path = access.path();
+        let local_path = self.safe_local_path(app_state, access);
+        match tokio::fs::metadata(&local_path)
+            .await
+            .and_then(|m| m.modified())
+        {
+            Ok(modified) => return Some(DateTime::<Utc>::from(modified)),
+            Err(e) if !is_path_missing_error(&e) => {
+                log::debug!(
+                    "Unable to read the modification time of {}: {e}",
+                    local_path.display()
+                );
+                return None;
+            }
+            Err(_) => {}
+        }
+        let db_fs = self.db_fs_queries.as_ref()?;
+        match db_fs.last_modified_in_db(app_state, path.as_ref()).await {
+            Ok(modified) => modified,
+            Err(e) => {
+                log::debug!(
+                    "Unable to read the modification time of {} from the database: {e:#}",
+                    path.display()
+                );
+                None
+            }
+        }
+    }
+
     pub(crate) async fn read_to_string(
         &self,
         app_state: &AppState,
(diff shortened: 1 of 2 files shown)


lovasoa:
Hi @ vaibhav8a ! The above looks like a raw llm output, feel free to edit it to explain your motivation in your own words, what you did and why, then switch the pr back to ready for review. You can explain why and how you used a coding agent, but I'm not accepting pull request that are not human reviewed.

PR #1428, opened 2026-09-06
refactor(render) :: drop the unread component name in handle_row

81reap:
`handle_row` looked up the current component's name and then only checked whether a component was open at all. The name was never read.

found by #1396


81reap:
CI failure is unrelated to this change, it comes from this commit on main https://github.com/sqlpage/SQLPage/commit/4c60e5d7884bbd5a3dc7ea8511270819d52ed7a6

lovasoa (approved):

lovasoa (approved):

PR #1429, opened 2026-09-06
fix(cookie) :: test the security attributes on a cookie

81reap:
Nothing checked that a cookie is `HttpOnly`, `Secure`, `SameSite=Strict` and scoped to `/` by default. Nothing checked that MySQL, MSSQL, and Oracle (which have no `boolean` literal) was able to set cookie overrides with `0`.

Adds tests for these cases.

found by #1396

### DIFF
--- a/tests/cookies/mod.rs
+++ b/tests/cookies/mod.rs
@@ -1,0 +1,31 @@
+use crate::common::req_path;
+use actix_web::http::header::SET_COOKIE;
+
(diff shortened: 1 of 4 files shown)


81reap:
CI failure is unrelated to this change, it comes from this commit on main https://github.com/sqlpage/SQLPage/commit/4c60e5d7884bbd5a3dc7ea8511270819d52ed7a6

lovasoa (approved):

PR #1434, opened 2026-09-07
fix(telemetry) :: render span fields through one implementation

81reap:
The span-field test and production function followed different implementations. Both now share one accumulator.

This change also allows testing that a span stack renders the nearest span first and that the first value wins for a given key.

discovered by #1396


lovasoa (approved):
thanks !

PR #1435, opened 2026-09-07
fix(http_client) :: test which variables select the system trust store

81reap:
`SSL_CERT_FILE` and `SSL_CERT_DIR` decide whether `sqlpage.fetch` and `send_mail` trust the system root certificates or the bundled ones. This addes tests for these cases.

discovered buy #1396


lovasoa (changes requested):
I don't think this is the right level to test this. This is verbose and does not actually test the feature. Certificate selection could be broken and this would pass

If you think this part of the code is brittle, I think a higher level test is more appropriate. One that tests end 2 end which certificates are selected when using fetch or send mail. This is the kind of integration tests we run inside ./examples. You could add an example about ssl certificates and a hurl test for it.

PR #1436, opened 2026-09-07
test(app_config) :: test the documented connection timeout defaults

81reap:
`configuration.md` documents that (1) the idle timeout defaults to 30 minutes, (2) the maximum lifetime to an hour, (3) that SQLite doesn't have those limits, and (4) that `0` disables the timeout. Adds tests for these cases.

discovered by #1396 

https://github.com/sqlpage/SQLPage/blob/6d32b4052e55e4824961f0b48fef2ca42b142298/configuration.md?plain=1#L18-L19

### DIFF
--- a/src/app_config.rs
+++ b/src/app_config.rs
@@ -863,6 +863,56 @@ mod test {
         assert!(error.contains("SMTP credentials require smtp_tls_mode"));
     }
 
+    fn config_with_timeouts(json: &str) -> AppConfig {
+        let mut config = serde_json::from_str::<AppConfig>(json).unwrap();
+        config.resolve_timeouts();
+        config
+    }
+
+    #[test]
+    fn a_zero_connection_timeout_disables_it() {
+        let config = config_with_timeouts(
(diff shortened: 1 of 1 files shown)


src/app_config.rs
```diff
@@ -863,6 +863,56 @@ mod test {
...
+        let config = config_with_timeouts(r#"{"database_url": "postgres://localhost/test"}"#);
+        assert_eq!(
+            config.database_connection_idle_timeout,
+            Some(Duration::from_mins(30))
+        );
+        assert_eq!(
+            config.database_connection_max_lifetime,
+            Some(Duration::from_hours(1))
+        );
+
```
lovasoa, line 899: you can add tests for the zero = None logic if you want. But we should not assert the default configuration values. These tests just add weight and have zero bug-catching power. They do not prevent the docs and actual values from drifting, they just add toil when changing defaults.

lovasoa (changes requested):
not useful imho

PR #1440, opened 2026-09-10
fix(chart): make xticks count tick positions

### LINKED ISSUE #733: Bug in the display of the chart component (x axis is not correctly aligned) reopened
### Introduction

The x axis shifts and is not aligned with the displayed value (cf screenshot)

### To Reproduce

in Postgresql

```sql
-- Create or replace the table
DROP TABLE IF EXISTS delete_table;
CREATE TABLE delete_table (
    x INTEGER NOT NULL,           -- Represents the hour
    y NUMERIC(10, 2) NOT NULL,    -- Represents the revenue at the hour
    series VARCHAR(50) NOT NULL   -- Represents the series (e.g., "last year", "this year")
);

-- Insert data into the table
INSERT INTO delete_table (x, y, series) VALUES
(10, 1.5, 'last year'),
(11, 1.7, 'last year'),
(10, 2.0, 'this year'),
(11, 1.9, 'this year'),
(12, 3.0, 'this year'),
(12, 2.5, 'last year'),
(13, 4.0, 'this year'),
(13, 3.5, 'last year'),
(14, 3.0, 'this year'),
(14, 7.5, 'last year'),
(15, 10.0, 'this year'),
(15, 1.5, 'last year'),
(16, 3.0, 'this year'),
(16, 5.5, 'last year'),
(17, 7.0, 'this year'),
(17, 8.5, 'last year'),
(18, 2.0, 'this year'),
(18, 2.5, 'last year'),
(19, 5.0, 'this year'),
(19, 3.5, 'last year'),
(20, 3.0, 'this year'),
(20, 6.5, 'last year'),
(21, 8.0, 'this year'),
(21, 3.5, 'last year');

-- chart component

SELECT 
    'chart' AS component,
    'bar' AS type,
    5 AS marker,
    True as labels,
    'Heures' AS xtitle,
    'Revenue (k€)' AS ytitle;

SELECT * from delete_table;
```

### Screenshots
here is the result 
image

### Expected behavior
cf screenshot the bars are not well aligned with the xaxis tick

### Version information

Postgresql

lovasoa:
## Summary

- translate numeric `xticks` values from requested tick positions to ApexCharts intervals
- assert the numeric chart fixture renders exactly the requested number of x-axis labels
- update the chart documentation and unreleased changelog entry

## Tests

- `npm run format`
- `npm test`
- `SQLPAGE_FIXTURE_BASE=http://127.0.0.1:18081 SQLPAGE_BINARY=/home/ophir/dev/SQLPage/target/debug/sqlpage npx playwright test fixtures/chart/test.ts --project=fixtures --reporter=line`

### DIFF
--- a/sqlpage/apexcharts.js
+++ b/sqlpage/apexcharts.js
@@ -393,9 +393,10 @@ sqlpage_chart = (() => {
       series,
     };
     if (labels) options.labels = labels;
-    // Numeric axes count intervals; category and time axes use tickAmount as a
-    // target for label density.
-    if (data.xticks) options.xaxis.tickAmount = data.xticks;
+    // ApexCharts counts numeric intervals rather than tick positions.
+    if (data.xticks)
+      options.xaxis.tickAmount =
+        xaxis_type === "numeric" ? Math.max(1, data.xticks - 1) : data.xticks;
     const chart = new ApexCharts(chartContainer, options);
     chart.render();
     if (window.charts) window.charts.push(chart);
--- a/sqlpage/apexcharts.js
+++ b/sqlpage/apexcharts.js
@@ -61,8 +61,17 @@ sqlpage_chart = (() => {
   /** @param {ChartSeries[]} series */
   const x_is_text = (series) => typeof series[0]?.data?.[0]?.x === "string";
 
-  /** @param {ChartSeries[]} series @param {string} chart_type */
-  function xaxis_type_for(series, chart_type, is_timeseries, is_horizontal) {
+  /**
+   * Numeric x values need an explicit axis type to retain their proportional
+   * spacing; otherwise ApexCharts treats them as evenly spaced categories.
+   *
+   * @param {ChartSeries[]} series
+   * @param {{chart_type:string, is_timeseries:boolean, is_horizontal:boolean}} options
+   */
+  function xaxis_type_for(
+    series,
+    { chart_type, is_timeseries, is_horizontal },
+  ) {
     if (is_timeseries) return "datetime";
     if (x_is_text(series)) return "category";
     if (
@@ -71,6 +80,18 @@ sqlpage_chart = (() => {
       NUMERIC_X_CHART_TYPES.includes(chart_type)
     )
       return "numeric";
+  }
+
+  /**
+   * ApexCharts expects intervals for numeric axes, while SQLPage exposes the
+   * more intuitive number of tick positions to users.
+   *
+   * @param {number|undefined} xticks
+   * @param {string|undefined} xaxis_type
+   */
+  function xaxis_tick_amount(xticks, xaxis_type) {
+    if (!xticks) return;
+    return xaxis_type === "numeric" ? Math.max(1, xticks - 1) : xticks;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


sqlpage/apexcharts.js
81reap, line 346: why not define `tickAmount` as part of the initial builder?

81reap, line 78: [nit] `x_is_text` is basically the same object being destructured

sqlpage/apexcharts.js
```diff
@@ -73,6 +82,18 @@ sqlpage_chart = (() => {
...
+  /**
+   * ApexCharts expects intervals for numeric axes, while SQLPage exposes the
+   * more intuitive number of tick positions to users.
+   *
+   * @param {number|undefined} xticks
+   * @param {string|undefined} xaxis_type
+   */
+  function xaxis_tick_amount(xticks, xaxis_type) {
+    if (!xticks) return;
+    return xaxis_type === "numeric" ? Math.max(1, xticks - 1) : xticks;
```
81reap, line 94: `numeric` is only possible if `is_horizontal` is `false` (see line 70 before or 79 after)

tests/js/chart_series.spec.ts
```diff
@@ -55,8 +70,22 @@ test("does not turn category-oriented charts into numeric axes", () => {
   const numeric = [series("a", { x: 1, y: 1 })];
```
81reap, line 70: adding a test for `[series("a", { x: "a", y: 1 })]` and checking `assert.equal(value_axis_labels(3, axis), 3);` instead of `undefined` may have caught the horizontal edge case

sqlpage/apexcharts.js
```diff
@@ -393,9 +413,7 @@ sqlpage_chart = (() => {
       series,
     };
     if (labels) options.labels = labels;
-    // Numeric axes count intervals; category and time axes use tickAmount as a
-    // target for label density.
-    if (data.xticks) options.xaxis.tickAmount = data.xticks;
+    options.xaxis.tickAmount = xaxis_tick_amount(data.xticks, xaxis_type);
```
81reap, line 416: the docs say this is tick interval not the number of ticks

> Number of Tick Intervals to show. Note: tickAmount doesn't affect datetime xaxis types. For numeric axes, use tickAmount: 'dataPoints' to match dataPoint counts.

https://apexcharts.com/docs/options/xaxis/

sqlpage/apexcharts.js
81reap: we may also not need the `- 1`

https://github.com/apexcharts/apexcharts.js/blob/cff2224488c1330bfeb2d72e30c1878e0a091847/src/modules/axes/AxesUtils.js#L124

81reap (changes requested):

sqlpage/apexcharts.js
lovasoa: I think we do need the -1, this is documented and I added a test for it https://apexcharts.com/docs/options/xaxis/


But the bigger concern is backwards compat, let's not release this in a patch release

PR #1445, opened 2026-09-11
Add server-side table sorting

### LINKED ISSUE #1442: Confusing sort interface
### Introduction

Any table which has more rows than the page size can deliver wrong results if sorted by the column headers in the table component. This is significant if you have a table lots of rows that is paginated by the pagination component.

### To Reproduce

The issue is very simple, just select a table with more tuples then your pagination (say a pagination of 100 for a total records of 500), sort it by the table header and then advance to the following page.

```sql
SELECT 'table' AS component, true AS sort,
    '/atti/edit.sql?id_soc={id}' AS edit_url,
    '/atti/delete.sql?id_soc={id}' AS delete_url;

SELECT
    t.id_atto::text AS _sqlpage_id,
    COALESCE(t.notaio::text, '—') AS "Notaio",
    COALESCE(t.filza::text, '—') AS "Filza",
    COALESCE(t.documento_n::text, '—') AS "Documento n.",
    to_char(t.data_stipula, 'DD/MM/YYYY') AS "Data Stipula",
    COALESCE(t.tipo_atto::text, '—') AS "Tipo Atto",
    COALESCE(t.regesto::text, '—') AS "Regesto"
FROM atti AS t
ORDER BY t.id_atto
LIMIT 100 OFFSET COALESCE(CAST(NULLIF($offset, '') AS INTEGER), 0);

SELECT 'pagination' AS component;
SELECT
    (g.i / 100 + 1) AS contents,
    sqlpage.link(sqlpage.path(), jsonb_strip_nulls(jsonb_build_object('offset', g.i, 'notaio', NULLIF($notaio, ''), 'filza', NULLIF($filza, ''), 'documento_n', NULLIF($documento_n, ''), 'sort', COALESCE(NULLIF($sort, ''), 'id_atto'), 'dir', NULLIF($dir, '')))) AS link,
    (g.i = COALESCE(CAST(NULLIF($offset, '') AS INTEGER), 0)) AS active
FROM generate_series(0, GREATEST((SELECT count(*) FROM atti AS t WHERE (NULLIF($notaio, '') IS NULL OR t.notaio::text ILIKE '%' || $notaio || '%') AND (NULLIF($filza, '') IS NULL OR t.filza::text ILIKE '%' || $filza || '%') AND (NULLIF($documento_n, '') IS NULL OR t.documento_n::text ILIKE '%' || $documento_n || '%')) - 1, 0), 100) AS g(i);
```

### Actual behavior

The order is valid only for the current page (and selection!). If I understand well, there is no way to forward it towards the database.
There is an open discussion "Table filtering, sorting, pagination." by [jeneizs-74](https://github.com/jeneizs-74) here:
https://github.com/sqlpage/SQLPage/discussions/778
It has no answer since Jan 15, 2025.

This may lead to errors and Is anyway confusing for the user who may think sorting shows the proper records, while he will get only the result of sorting the page records.

### Expected behavior

I think sorting should happen on the entire table. At least there should be some way of getting sorting column(s) and direction so to force a new select on the backend.

lovasoa:
Closes #1442

## Summary
- add opt-in `server_sort_column` table headers that reload with `sort_<column>=ASCENDING|DESCENDING`
- preserve existing in-browser `sort` behavior for other columns
- document the property and use it in the simple website example

## Testing
- `npm test`
- `npx playwright test fixtures/table-server-sort/test.ts --project=fixtures --reporter=line`
- `hurl --test --error-format long examples/simple-website-example/test.hurl`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG.md
 
+## Unreleased
+
+- The table component supports `server_sort_column`. Columns named by this property reload the current page with a `sort_<column>=ASCENDING|DESCENDING` URL parameter when their headers are clicked, enabling database-side sorting before pagination.
+
 ## v0.46.1
 
 - Fixed a regression introduced in v0.46 that could replace a variable with `NULL` while building a value that also used database expressions and `sqlpage.*` functions. For example, this API request could lose `john.doe` and produce a URL ending at `https://api.example.com/`:
--- a/examples/official-site/sqlpage/migrations/01_documentation.sql
+++ b/examples/official-site/sqlpage/migrations/01_documentation.sql
@@ -949,7 +949,8 @@ Advanced users can apply custom styles to table columns using a CSS class with t
 
 INSERT INTO parameter(component, name, description, type, top_level, optional) SELECT 'table', * FROM (VALUES
     -- top level
-    ('sort', 'Make the columns clickable to let the user sort by the value contained in the column.', 'BOOLEAN', TRUE, TRUE),
+    ('sort', 'Make the columns clickable to let the user sort by the value contained in the column in the browser.', 'BOOLEAN', TRUE, TRUE),
+    ('server_sort_column', 'A column name, or JSON array of column names, whose headers reload the current page with a sort_<column>=ASCENDING or sort_<column>=DESCENDING URL parameter. Use this for database-side sorting, particularly with pagination. Your SQL query must read the parameter and apply the corresponding ORDER BY safely.', 'JSON', TRUE, TRUE),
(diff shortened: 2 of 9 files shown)


cbertelli:
I don't think that it is necessary to limit sorting to the indexed columns. The user may decide to wait. It would be very useful to point out the absence of an index, but sorting should be effective and allowed anyway.
The problem may arise when a column is decoded by an alphanumeric label. The user will not expect the column to be sorted by the integer id or even an alphanumeric code, he will expect it to be sorted by label. This is indexed, sure, but references another table and the interface is asked to show a sorted label, not a sorted code/id (I hope it's not too obscure).

PR #1447, opened 2026-09-11
Fix chart tooltip links on hover

lovasoa:
## Summary

- Keep linked chart tooltips open while moving from a data point to the tooltip.
- Close the tooltip after leaving it.
- Cover Gantt, bar, and scatter linked-tooltip handoffs.

## Testing

- `npm test`
- `SQLPAGE_FIXTURE_BASE=http://127.0.0.1:18081 SQLPAGE_TEST_BASE=http://127.0.0.1:18081 npx playwright test fixtures/chart/test.ts --project=fixtures --grep 'interactive data point link|keeps a data point link open'`\n\nThe line fixture does not currently surface its tooltip link, and the pie fixture places its linked tooltip outside the viewport; those are separate from the hover-close issue.

### DIFF
--- a/sqlpage/apexcharts.js
+++ b/sqlpage/apexcharts.js
@@ -419,9 +419,10 @@ sqlpage_chart = (() => {
     tooltip.className = "apexcharts-tooltip-text";
     tooltip.style.fontFamily = "inherit";
 
-    const seriesName = document.createElement("div");
+    const seriesName = document.createElement(link ? "a" : "div");
     seriesName.className = "apexcharts-tooltip-y-group";
     seriesName.style.fontWeight = "bold";
+    if (seriesName instanceof HTMLAnchorElement) seriesName.href = link;
     seriesName.innerText = name;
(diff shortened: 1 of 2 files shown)


81reap (commented):
so this does work, but it feels very inconsistent. to save ourselves the trouble we should consider using a browser native API like https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover#hint

it's supported everywhere but Safari, where its available in pre-release https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover#browser_compatibility


maybe even an upstream fix to apexcharts? seems like this is a repeated issue
- https://github.com/apexcharts/apexcharts.js/issues/639
- https://github.com/apexcharts/apexcharts.js/issues/2794

lovasoa:
Yeah, that's hacky. I'd rather fix this in apexcharts itself: https://github.com/apexcharts/apexcharts.js/pull/5307

Issue #1448, opened 2026-09-11
Move to a better type system


81reap:
Currently the front-end is a bunch of JavaScript files typed with JSDoc which has limitations such as ::
1. Doesn't allow for new features to be added easily https://github.com/jsdoc/jsdoc/pull/2139
2. Doesn't support negative numbers https://github.com/jsdoc/jsdoc/issues/2131 and a fix has been open for a while now https://github.com/hegemonic/catharsis/pull/76

If we switch the frontend over to Typescript, it will cost us a build step but we gain ::
1. shared types between test and prod
2. being able to import dependency types like ApexCharts
3. having a build system handle compression and optimizations, so maintainers can freely write code in inefficient but more readable ways
4. moving the front end 3rd party dependices out of the rust build step and into the typescript build step

lovasoa:
I am 100% onboard with this. It's past time we get both a better developer experience, and a more solid frontend.

I think the first step towards that goal is to introduce a build + package step, while keeping as much of the code unchanged as possible. Ideally after this first step we should be able to run things like `npm run dev -- examples/official-site` to work on SQLPage with live reload, `npm test` to run the full test suite (rust + frontend) with unified and minimal reporting (very useful for agents), `npm run build` to build everything. Maybe we could switch to a unified tool like https://rsbuild.rs/ which could also replace biome. The goal for this first step would be to improve the build infrastructure, and maybe even to delete more lines of code than we introduce (the build system would what we do manually in build.rs).

Then once we have that, open a second pr with the typescript switch, that almost does not touch the build pipeline, just adds annotations to the frontend code. We could even strive to make this a PR with a net negative line count, replacing jsdoc annotations with inline ts annotations. 

AI agents can do most of the migration, but from experience, they have a tendency to very quickly overbloat this kind of projects, trying to keep backwards compatibility in places that do not make sense, and ending with overcomplicated designs. If you do it, please keep your agent on a tight leash, maybe even with an explicit goal of ending on a diff that is a net negative line of code count, even if that needs making compromises.

lovasoa:
Some ideas for next steps that a good build infrastructure would allow:
 - always up to date documentation:
   - I'd love our `configuration.md` to move to a page on the official site that is autogenerated from the rust configuration datastructure, so that it's always up to date
   - documenting components directly from comments at the top of handlebars files instead of inside .sql migrations
   - documenting sqlpage functions directly from the rust source code
[...]

PR #1449, opened 2026-09-12
fix(chart) :: group points into series without `Object.prototype`

81reap:
Points were grouped in a plain object keyed by the `series` value. A series named `toString` or `constructor` threw, rendering nothing. Numeric series names were enumerated in ascending order ahead of text ones rather than keeping the order the rows came in.

### DIFF
--- a/sqlpage/apexcharts.js
+++ b/sqlpage/apexcharts.js
@@ -53,7 +53,7 @@ sqlpage_chart = (() => {
   /** @typedef {number|string|Date} XValue */
   /** @typedef { {x:XValue, y:number|null, z?:number, fillColor?:string} } ChartPoint */
   /** @typedef { {name:string, data:ChartPoint[]} } ChartSeries */
-  /** @typedef { { [name:string]: ChartSeries } } Series */
+  /** @typedef { Map<string, ChartSeries> } Series */
 
   /** @param {XValue} x @returns {number|string} equal x values share a key */
   const x_key = (x) => (x instanceof Date ? x.getTime() : x);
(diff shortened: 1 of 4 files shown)


lovasoa:
I agree the generated code is cleaner, but I'm not sure I understand the bug. Is there a real bug ? 

If the only way to reproduce the "bug" is with somthing like 

```sql
SELECT 'chart' AS component, 'constructor' AS type; -- constructor is not a valid chart type, this is documented
WITH points(series, x, y, color) AS (VALUES ('A', 'Q2', 2, 'toString')) SELECT * FROM points; -- toString is not a valid color, documented too
```

then I don't think this is a bug at all. You can rename it to a refactor or a code cleanup, and not mention it in the changelog at all.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+- Fixed charts misreading a `type` or a `color` named after a built-in JavaScript property, such as `constructor` or `toString`.
```
lovasoa: let's just remove this. Unclear and confusing to real sqlpage users. Nothing changes for them

lovasoa (changes requested):

CHANGELOG.md
81reap: removed

81reap:
This was discovered as a side effect of #1448 but it didn't make sense to belong in that stack. Your right tho that this PR is doing too much and not all of this is an SQLPage issue. I have upstreamed https://github.com/apexcharts/apexcharts.js/pull/5308 and https://github.com/apexcharts/apexcharts.js/pull/5309 as well as updated this PR to focus only on the bug in SQLPage as well as updated the examples to be more clear.

lovasoa:
Good job!

lovasoa (approved):

PR #1452, opened 2026-09-14
fix(big_number) :: preserve color and spacing

### LINKED ISSUE #1451: 0.46.2 - big number bar COLOR attribute does not set color of the text
### Introduction

Till 0.46.1, setting a color for big number bar card used to set background color of the card as well as the text color. Once I changed the binary to 0.46.2, color of the text is always black  in light mode and always white in dark mode.

### To Reproduce

Just change the binary from 0.46.1 to 0.46.2. Check with example 2 in big number bar component documentation. I see that the documentation shows it with the same black/white color only - I don't recall whether it was like that in previous version's documentation. Since documentation is also served by sql-page, it could show the latest behavior.

Actual SQL I am using is at https://github.com/vsbabu/xp/blob/main/r_bignumber_bar.sql

Note that the progress color is showing up fine as given in the sql.

### Actual behavior

There are no errors in the command line or on web developer console. The color determination logic for the text as shown in web developer console is exactly similar in both cases, just that the color rendered is wrong.

I also noticed that the margins or paddings have also increased a bit. Both version's screenshots are given below.

### Screenshots

Image

### Expected behavior
Nice to have it like in 0.46.1

### Version information
0.46.2
 - OS:
 - Database [e.g. SQLite, Postgres]
 - SQLPage Version [found when hovering the default footer of pages]:

lovasoa:
## Summary

Tabler 1.5 stopped headings from inheriting the `bg-*-lt` foreground and increased default card vertical padding from `1rem` to `1.25rem`. This PR makes the big-number value color explicit, keeps linked values the same color, and restores compact spacing with a component-scoped variable.

This went unnoticed because the template did not change, documentation renders only the current behavior, and no browser test covered computed big-number styles. The new Playwright fixture checks linked and unlinked values in light and dark themes without pinning palette RGB values.

I audited the other `bg-*-lt` template uses: timeline icons still inherit correctly, table row color is documented as background-only, and the Tabler 1.5 modal regression was already fixed separately.

## Screenshots

Same SQL and 900x900 viewport in each image; light theme is left, dark theme is right.

**v0.46.1**

v0.46.1 light and dark

**v0.46.2**

v0.46.2 light and dark

**This PR**

PR light and dark

## Tests

- `npm test`
- `npx playwright test fixtures/big-number/test.ts --project=fixtures --reporter=line`

Closes #1451


kryskool (approved):

tests/end-to-end/fixtures/big-number/test.ts
```diff
@@ -0,0 +1,34 @@
+import { expect, test } from "../../fixture";
+
+for (const theme of ["light", "dark"]) {
+  test(`colored values match their cards in the ${theme} theme`, async ({
+    page,
+  }) => {
+    await page.locator("html").evaluate((html, selectedTheme) => {
+      html.dataset.bsTheme = selectedTheme;
+    }, theme);
```
81reap, line 9: we set theme on `body` not `html`

https://github.com/sqlpage/SQLPage/blob/2cb3d02b765dfaf427943bad1894f7d512eb053b/examples/CRUD%20-%20Authentication/sqlpage/templates/shell.handlebars#L66

https://github.com/sqlpage/SQLPage/blob/2cb3d02b765dfaf427943bad1894f7d512eb053b/examples/light-dark-toggle/test.hurl#L21

81reap (approved):

tests/end-to-end/fixtures/big-number/test.ts
lovasoa: thanks, fixed!

PR #1454, opened 2026-09-15
Media

olivierauverlot:
A HTML5 media component for playing video or audio files.


lovasoa:
Maybe this should be integrated into the existing https://sql-page.com/component.sql?component=carousel ?

olivierauverlot:
That doesn’t seem like a good idea to me. The media component has many options, which would make using a single component more complicated. Moreover, using a carousel component to play a sound file doesn’t seem intuitive to me.

PR #1456, opened 2026-09-16
fix(build) :: import 3rd party frontend assets from a list

### LINKED ISSUE #1463: Split the build
In order to accomplish #1448 we need to split `build.rs` so that we have npm responsible for the frontend and cargo be responsible for the backend.

81reap:
Before assets were build by scanning for `/* !include https://... */` comments to fetch the asset from a CDN.

Now the assets are a list in `build.rs`.

This ends up fixing another bug where `DOC_RS` would return in the middle of a scan on the first include which would leave the gzip stream unfinished.

---


lovasoa (changes requested):
Hi! Can you please organize the PRs so that all are small enough to be reviewed, yet none breaks the project?

81reap:
@lovasoa not sure how to do both here. I chose to split up this and #1457 as the diff was getting quite big.

That said, with GitHub stacks, we can review both PRs and merge them together so that project is never in a broken state!

lovasoa:
It looks like the next pr contains changes unrelated to what you are doing here. 

I think you could make a good no-brainer pr with net negative lines of code by keeping the exact exact same output files, but removing the include comment mechanism, hardcoding the concatenation logic in build.rs and copying from node modules instead of the web.

81reap:
I originally chose not to do it that way because it required refactoring `build.rs` in almost every commit in the stack and temporarily importing `node_modules` into the docker image. I didn't think that would be helpful to the review process. 

Pushed up a new stack tho.

lovasoa (approved):
thanks 👍

PR #1457, opened 2026-09-16
fix(build) :: import 3rd party frontend assets from npm

### LINKED ISSUE #1463: Split the build

81reap:
`build.rs` downloaded Tabler, ApexCharts, Tom Select and the Tabler icon sprite from a CDN on every clean build, with its own retry loop and a cache.

Now libraries are installed from npm and improted from `node_modules`. `cargo install sqlpage` and `docs.rs` don't need network anymore, but do need `npm ci` to run first.

---


.github/workflows/ci.yml
```diff
@@ -161,6 +180,11 @@ jobs:
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@v7
+      - uses: actions/setup-node@v7
+        with:
+          node-version: 26
+          cache: 'npm'
+      - run: npm ci --omit=dev --ignore-scripts
```
lovasoa: can we factor that so that we don't have to repeat the precise versions n times ?

```
 - uses: ./.github/actions/install-frontend-dependencies
```

81reap: yes, actually I was planning on fixing this with nix, but that is blocked by the typescript stack, which is blocked by the current NPM stack. ofc this list isn't final and will change for the better as we edit, merge, and talk over commit issues :) 

Screenshot_2026-09-18_09-01-37

81reap: ah actually I see that you already did it. thanks!

PR #1470, opened 2026-09-20
fix(build) :: refactor handling browser assets to npm

81reap:
Before `build.rs` used to be responsible for compressing assets with gzip and rewrite file names to be cache bustable.

Now we use standard node libraries to handle all browser assets in the frontend npm build. `build.rs` still needs to handle `ICON_SPRITE`.

---


lovasoa:
Follow-up on the build-complexity review: I pushed a one-command checkout build to this PR branch (`npm run build:rust`) and updated `CONTRIBUTING.md` to show the frontend build prerequisite, release/ODBC flags, and the faster `cargo build` path after assets exist.

I kept `npm ci` out of `build.rs`: the published crate contains prebuilt assets but not the npm project, and Cargo rebuilds should not reinstall `node_modules`. The existing `build.rs` missing-sprite error already tells contributors to run `npm ci && npm run build`.

Commits: 286f55b and 0fd75c4. I verified both updated files by reading them back from the branch; full CI is still pending.

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "scripts": {
     "build": "node scripts/build-frontend.mjs",
+    "build:rust": "npm ci --ignore-scripts && npm run build && cargo build",
```
81reap, line 6: this is good as a stop gap, but IMO `npm build:rust` being responsible for building the app doesn't seem the like the right pattern for a mostly rust codebase. I think a better path forward would be to add a `flake.nix` and use `nix flake check` for running everything, `nix fmt` for formatting across languages, `nix run .#NAME` for individual steps of the ci. This way we are also simplifying the CI to rely on the generic nix command

ps :: I was having issues merging a whole stack at once yesterday (something with the github backend) so we may need to merge this one by one to avoid that if its still an issue

pps :: I'm a fan of reabse and merge but up to you if you still wanna squash the commits

81reap: @lovasoa these are my thoughts but I'm aligned on the better documentation. I can start raising the Typescript stack and `flake.nix` stack in parallel after we merge this one

lovasoa: I am not a huge fan of adding one more dependency to build.

IMHO the nix configuration should live and be maintained in https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/sq/sqlpage/package.nix

81reap: so nix can be a lot of things depending on how you use it. I would compare `flake.nix` similar to a make file in this case. you can still build and run SQLPage without it by running `npm build` and `cargo build`. 

https://nix.dev/manual/nix/2.18/command-ref/new-cli/nix3-develop

the benefit is that when you run `nix develop` you get dropped into a shell with all of the deps needed to build SQLPage, so when you build locally it builds with the exact same dependencies and versions as the CI would. Like a make file, it also allows you to have one command that maps to multiple commands

That's why in a previous comment I had mentioned that I was working on upstreaming the missing OBDC clients so that we could use it in our dev shell

https://github.com/NixOS/nixpkgs/pull/552148


I agree the nix package config needs to live in the nixpkgs repo otherwise their HydraCI will not pick it up and build into their cache

lovasoa (approved):

package.json
81reap: @lovasoa I created https://github.com/sqlpage/SQLPage/issues/1483 so that we can continue this discussion without blocking the merge for this PR. I will update the issue with more details and examples once I get home from work

